PR #92, opened 2015-02-14
Fix scaling of svg.

schmmd:


schmmd:
@codeviking here's my fix.  I think it happened to work with your monitor resolution earlier.

PR #93, opened 2015-02-14
Overlay a box containing outputs.

schmmd:


schmmd:
@codeviking I coded up the functionality to provide the outputs overlay.  I'm concerned however that the outputs box will overlap the graph often.  I can think of a few solutions:

* The outputs could take up the left area of the graph.  The graph would then be centered in the remaining area.
* The outputs could be to the left of the graph box.
* Somehow magically the graph would never overlap the outputs box.

I figure you might have a better idea.  I know you are busy and no worries if you can't get to this right away.  But if you have any CSS magic up your sleeve, feel free to push to this branch.

image

codeviking:
This looks good.

How many outputs are there usually?  A lot?  If so we should think about having a scrollable side column (as you suggested), thus ensuring that the graph and outputs list don't overlap.   I could whip that up real quick like.

schmmd:
@codeviking here are some examples:

* [Full pipeline run](http://s3-us-west-2.amazonaws.com/ai2-s2/pipeline/output/publish/summary/build-index-spark-2015-01-12-21-12-39.html)
* [A single step with evaluations](http://ai2-s2.s3.amazonaws.com/pipeline/output/publish/summary/pdfjs-2015-02-09-17-41-35.html)
* http://ai2-s2.s3.amazonaws.com/pipeline/output/publish/summary/MeasureMetadata-2015-02-12-10-55-09.html

Often there does end up being space in the upper-left.

PR #96, opened 2015-02-18
link fixup

cristipp:
Fixup the workflow output such that s3n links are mapped into clickable http links. See here an example where this is broken: http://ai2-s2.s3.amazonaws.com/pipeline/output/publish/summary/commoncrawl-2015-02-18-22-57-47.html.

### DIFF
--- a/pipeline/src/main/scala/org/allenai/pipeline/Workflow.scala
+++ b/pipeline/src/main/scala/org/allenai/pipeline/Workflow.scala
@@ -69,8 +69,10 @@ object Workflow {
   }
 
   private def link(uri: URI) = uri.getScheme match {
-    case "s3" => new java.net.URI("http", s"${uri.getHost}.s3.amazonaws.com", uri.getPath,
-      null).toString
+    case "s3" | "s3n" =>
+      new java.net.URI("http", s"${uri.getHost}.s3.amazonaws.com", uri.getPath, null).toString
+    case "file" =>
+      new java.net.URI(null, null, uri.getPath, null)
     case _ => uri.toString
   }
 


jakemannix:
@schmmd, assigning to you because you said this was messy, and hence, you know what to do. :)

schmmd:
@cristipp I'm not sure this fixes the links.  The root problem is that the outputs in your example are directories which are not viewable in S3.  @rodneykinney is aware of this problem.  We briefly considered some work-arounds last week, but gave them up as they were too much of a pain.

I'm OK with the change as it's proposed, but it doesn't buy us too much.  Assigning back to @cristipp so he can merge if he please.

cristipp:
I'll submit this, as at least gets us an inch closer to the intent. But I won't fret about deploying it.

I do have a couple goals:
* make it super easy to read (as a human) the outputs of a pipeline stage. 
* regardless of the on-disk storage format.
It is quite unsatisfying to not be able to do it with a simple click. There are at least two easy-ish solutions:

A. Spark shell. Slice and dice your data with the full power of Spark. We do have some rough APIs to read the data we wrote, so it's workable. An example here: https://github.com/allenai/s2-offline/blob/master/pipeline/src/main/scala/org/allenai/scholar/pipeline/spark/imports/CiteseerxQuery.scala#L29. I'm going to poke at this a bit. Alas, this is still quite heavy weight.

B. s3proxy. It should be fairly straightforward to write a personal server that uses one's credentials and translated localhost:12345 URLs into s3 SDK requests and back. Making a shared server is scarier, as I'm not sure how 3rd party HTTP auth works in AWS world (if at all), but perhaps we have more knowledge about this that I'm not aware of?

On the web, I could not find anything better than signed URLs for CloudFront, which seem quite inconvenient as well.

PR #97, opened 2015-02-20
Use SS more sparingly to help git grep.

schmmd:


schmmd:
@rodneykinney, a comprimise

pipeline/src/main/scala/org/allenai/pipeline/ColumnFormats.scala
```diff
@@ -13,8 +13,11 @@ import java.util.regex.Pattern
 trait ColumnFormats {
   private type SS[T] = StringSerializable[T]
```
jkinkead: Move to right before use in the implicit defs?

rodneykinney:
I'm fine with getting rid of the alias altogether.  I only introduced it as a workaround, because scalariform would reformat the lines to exceed the line length of the style checker.

schmmd:
@rodneykinney if you run into that problem (scalariform making your lines too long) let me or Jesse know.  There's a few simple tricks to learn about how to split lines of code.

jkinkead:
The newer scalariform fork we started using in November is *much* easier to convince to keep lines short, since the default behavior for multi-line `defs` is to split it out further, rather than joining all on one line.

PR #99, opened 2015-02-26
Pipeline code documentation and cleanup.

schmmd:


schmmd:
I still wonder about https://github.com/allenai/common/blob/docs/pipeline/src/main/scala/org/allenai/pipeline/IoHelpers.scala#L17.  It seems weird for serialization helper code to be in an `IoHelpers` object.

rodneykinney:
I had spray-json in mind.  I like how I can import `DefaultJsonProtocol._` and it has everything I ever want.  Things are split into different namespaces so you can choose what you want, but in practice, I only  ever import the one top-level object that includes everything.

schmmd:
Fair enough.

PR #102, opened 2015-03-07
Split pipeline sub-project into https://github.com/allenai/pipeline

rodneykinney:
@cristipp @jakemannix FYI.  I've update s2-offline to depend on the new binary.


build.sbt
```diff
@@ -39,13 +39,7 @@ lazy val webapp = Project(
   settings = buildSettings
 ).enablePlugins(LibraryPlugin).dependsOn(core, testkit % "test->compile")
 
-lazy val pipeline = Project(
```
cristipp: I'd expect you already have the https://github.com/allenai/pipeline repo. If you have that, then I'd expect this PR to replace the deleted code with a new Maven-based jar. So everything keeps working...

schmmd:
This is great.  It's clear the pipeline is not a good solution going forward.

[Just kidding](https://github.com/allenai/pipeline)!

PR #103, opened 2015-03-07
Adds Iterator.parForeach and Iterator.parMap

dirkgr:
@schmmd 

In the course of this, I learned about `blocking { ... }`, which is amazing for preventing deadlocks. Check out http://blog.jessitron.com/2014/02/scala-global-executioncontext-makes.html.

### DIFF
--- a/core/src/main/scala/org/allenai/common/ParIterator.scala
+++ b/core/src/main/scala/org/allenai/common/ParIterator.scala
@@ -1,37 +1,68 @@
 package org.allenai.common
 
 import java.util.concurrent.Semaphore
+import java.util.concurrent.atomic.AtomicReference
 
 import scala.concurrent._
 import scala.concurrent.duration.Duration
+import scala.util.control.NonFatal
 
 object ParIterator {
   val queueLimit = 1024
+  val exceptionLimit = 1024
 
-  implicit class ParIteratorEnrichment[T](val input: TraversableOnce[T]) extends AnyVal {
+  implicit class ParIteratorEnrichment[T](val input: Iterator[T]) extends AnyVal {
 
     def parForeach(f: T => Unit)(implicit ec: ExecutionContext): Unit = {
       // If there are a billion items in the iterator, we don't want to create a billion futures,
       // so we limit the number of futures we create with this semaphore.
       val sema = new Semaphore(queueLimit)
-      input.foreach { item =>
-        // Try to pass it off to a future. If no futures are available, do the work
-        // in this thread.
-        val success = sema.tryAcquire()
-        if (success) {
-          Future {
-            try {
-              f(item)
-            } finally {
-              sema.release()
-            }
+
+      val firstException = new AtomicReference[Option[(Int, Throwable)]](None)
+      def setException(index: Int, exception: Throwable): Unit = {
+        var success = false
+        while (!success) {
+          val current = firstException.get()
+          current match {
+            case None =>
+              success = firstException.compareAndSet(None, Some((index, exception)))
+            case Some((oldIndex: Int, _)) if index < oldIndex =>
+              success = firstException.compareAndSet(current, Some((index, exception)))
+            case _ =>
+              success = true
           }
-        } else {
-          f(item)
         }
       }
+
+      input.zipWithIndex foreach {
+        case (item, index) =>
+          // Try to pass it off to a future. If no futures are available, do the work
+          // in this thread.
+          val success = sema.tryAcquire()
+          if (success) {
+            Future {
+              try {
+                f(item)
+              } catch {
+                case NonFatal(e) => setException(index, e)
+              } finally {
+                sema.release()
+              }
+            }
+          } else {
+            try {
+              f(item)
+            } catch {
+              case NonFatal(e) => setException(index, e)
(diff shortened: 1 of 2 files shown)


dirkgr:
I should probably mention the tests take a long time (2m or so), because I need to go over the queue limit to test that code path. If that's a problem, I'll have to think of something.

core/src/main/scala/org/allenai/common/ParIterator.scala
```diff
@@ -0,0 +1,58 @@
...
+      // so we limit the number of futures we create with this semaphore.
+      val sema = new Semaphore(queueLimit)
+      input.foreach { item =>
+        // Try to pass it off to a future. If no futures are available, do the work
+        // in this thread.
+        val success = sema.tryAcquire()
+        if (success)
+          Future {
+            f(item)
+            sema.release()
```
jkinkead: Should be in a finally, yes?

dirkgr: Yes, of course.

schmmd:
Please ignore the test case by default if it takes > 2 minutes.  Add a comment to warn people changing the methods.

core/src/main/scala/org/allenai/common/ParIterator.scala
```diff
@@ -0,0 +1,58 @@
+package org.allenai.common
+
+import java.util.concurrent.Semaphore
+
+import scala.concurrent._
+import scala.concurrent.duration.Duration
+
+object ParIterator {
+  val queueLimit = 1024
```
schmmd: Why not make this a configurable parameter with a default?

core/src/main/scala/org/allenai/common/ParIterator.scala
```diff
@@ -0,0 +1,58 @@
...
+            sema.release()
+          }
+        else
+          f(item)
+      }
+      // wait for all threads to be done
+      blocking { sema.acquire(queueLimit) }
+    }
+
+    def parMap[O](f: T => O)(implicit ec: ExecutionContext): TraversableOnce[O] = new Iterator[O] {
```
schmmd: Please add a scaladoc for this method (is order preserved, etc.)

core/src/main/scala/org/allenai/common/ParIterator.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import scala.concurrent._
+import scala.concurrent.duration.Duration
+
+object ParIterator {
+  val queueLimit = 1024
+
+  implicit class ParIteratorEnrichment[T](val input: TraversableOnce[T]) extends AnyVal {
+
+    def parForeach(f: T => Unit)(implicit ec: ExecutionContext): Unit = {
```
schmmd: Please add a scaladoc.

schmmd:
Do you prefer `parForeach` to `parallelForeach`?  Since `par` is already in Scala I guess it's OK.

schmmd:
Please don't introduce style errors or warnings into common.  The shippable log lists both.

core/src/main/scala/org/allenai/common/ParIterator.scala
schmmd: I think we should make this a configurable parameter.  Otherwise the client has no recourse if they want to use this library and performance tune.

core/src/main/scala/org/allenai/common/ParIterator.scala
dirkgr: I did this just now, but this points to a much bigger problem. If an exception happens in `f`, I have to import that exception to the calling thread and throw it there. In other words, I have to call `get` on each of those futures, which means I'll have to change the whole way `parForeach` works.

dirkgr:
I reduced all the test timings by 90%, and now it runs reasonably fast. If your machine is super loaded, the tests could fail, because they depend on timing, and there is now less of a margin of error. I think it's acceptable, and preferable to ignoring tests.

schmmd:
Sounds good for now.

dirkgr:
I think that's all concern addressed.

dirkgr:
Good to merge?

dirkgr:
Sorry about the style issues. I started trusting the hooks too soon I guess.

schmmd:
hooks = scalariform

There's also the dreaded evil ScalaStyle!

dirkgr:
I thought the hook would not let me commit unless I meet the style guidelines?

Cheers!

Dirk

On March 10, 2015 at 11:26:56, Michael Schmitz (notifications@github.com) wrote:

hooks = scalariform

There's also the dreaded evil ScalaStyle!

schmmd:
They were warnings.  Shippable will fail on errors.

dirkgr:
I think I addressed everything we talked about over lunch.

schmmd:
I like how it turned out.  Assigning back to you since you keep having more ideas...

dirkgr:
Do you think this is worthy of a quick note during the engineering meeting, so people know it's there? I wouldn't know to look for this otherwise.

schmmd:
That would be great.  This week we have time for 10 minutes.

PR #107, opened 2015-04-01
Adds startup timestamp to the info route

dirkgr:
@jkinkead, can you take a look?

When this is deployed, I'll change the daily eval script to use this to detect when QI has restarted, and disregard any evaluation that was affected.

### DIFF
--- a/webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
+++ b/webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
@@ -15,7 +15,7 @@ import spray.routing.Route
   * @param info the info to serve
   */
 class InfoRoute(val info: Map[String, String] = Map.empty) {
-  val infoWithTime = info.updated("startupTimestamp", (System.currentTimeMillis() / 1000L).toString)
+  val infoWithTime = info.updated("startupTimestamp", System.currentTimeMillis().toString)
 
   def withVersion(version: Version): InfoRoute = {
     new InfoRoute(
--- a/webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
+++ b/webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
@@ -15,6 +15,8 @@ import spray.routing.Route
   * @param info the info to serve
   */
 class InfoRoute(val info: Map[String, String] = Map.empty) {
+  val infoWithTime = info.updated("startupTimestamp", (System.currentTimeMillis() / 1000L).toString)
+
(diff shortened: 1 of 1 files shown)


webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
```diff
@@ -15,6 +15,8 @@ import spray.routing.Route
   * @param info the info to serve
   */
 class InfoRoute(val info: Map[String, String] = Map.empty) {
+  val infoWithTime = info.updated("startupTimestamp", (System.currentTimeMillis() / 1000L).toString)
```
jkinkead: You should keep this millis to match the git timestamp.

webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
```diff
@@ -37,14 +39,14 @@ class InfoRoute(val info: Map[String, String] = Map.empty) {
       pathEndOrSingleSlash {
         respondWithMediaType(MediaTypes.`application/json`) {
           complete {
-            info.toJson.prettyPrint
+            infoWithTime.toJson.prettyPrint
```
jkinkead: It's a little odd that this copy is used in place of `info` in some places, but not everywhere.

I'd prefer if this were added in the same pattern as `withName` and `withVersion` - like `withStartupTime` (either taking the startup time value to use, or else no-args & capturing the current time).

dirkgr: I don't like that because then you have to call it explicitly when you use it. But I don't dislike it very strongly, so I made the change anyways.

dirkgr:
The real fix to this issue should have been to fix all the random errors we have in every evaluation, or at least make them warnings, and then fail the evaluation when there is a real error.

@codeviking, and chance we can get that in EF2.0?

codeviking:
It's in spec to fail loudly if there's an exception encountered during execution, but I'm not sure what "fix all the random errors we have in every evaluation" means.

webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
jkinkead: You could also just add it to the `withVersion` implementation (maybe renaming to `withVersionAndStartupTime` if you're feeling verbose).

jkinkead:
LGTM - feel free to move the timestamp to the `withVersion` call if you like; else, merge!

PR #108, opened 2015-04-23
Enable caching properly.

schmmd:


jkinkead:
LGTM!

shippable.yml
```diff
@@ -3,11 +3,11 @@ scala:
 - 2.11.5
 jdk:
 - openjdk7
```
jkinkead: I just noticed that this is using 7. Maybe update in a future PR?

PR #111, opened 2015-04-28
Publish using Shippable.

schmmd:


schmmd:
@jkinkead this should successfully make common publish via shippable.  We could merge this pull request or wait until I integrate with sbt-plugins.  Please take a look and review in any case!

admin/publish-release.sh
```diff
@@ -0,0 +1,55 @@
...
+# Publish to BinTray if the HEAD commit is tagged with a version number.
+if [ "$PULL_REQUEST" != "false" ]; then
+  echo "Shippable is building a pull request, not publishing."
+  echo "PULL_REQUEST is equal to $PULL_REQUEST"
+  exit 0
+fi
+
+if [ "$BRANCH" != "master" ]; then
+  echo "Shippable is building on branch $BRANCH, not publishing."
+  echo "BRANCH is equal to $BRANCH"
```
jkinkead: Remove? This is already displayed above.

admin/publish-release.sh
```diff
@@ -0,0 +1,55 @@
...
+  exit 0
+fi
+
+numParents=`git log --pretty=%P -n 1 | wc -w | xargs`
```
jkinkead: A comment on this would be helpful (mostly the `xargs` hack). I think `tr -d ' '` might work, too.

admin/publish-release.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ $numParents -ne 2 ]; then
+  echo "$numParents parent commits of HEAD when exactly 2 expected, not publishing."
+  exit 0
+fi
+
+# Shippable only runs a single build when a pull request is merged, so we need to list
+# all tags from the merged commits.
+firstMergedCommit=`git rev-list HEAD^2 --not HEAD^1 | tail -n 1`
```
jkinkead: I can't get the `^2` revision working reliably on my git branches. It may not always point to something. I'm actually having a hard time understanding what it even means.

I *think* that `git rev-list HEAD~1..HEAD` is what you want (note that `HEAD~1..HEAD` is the same as `HEAD --not HEAD~1`, but is a little more readable).

FWIW you could add `--reverse --max-count=1` to get the `tail -n 1` effect.

build.sbt
```diff
@@ -19,27 +19,30 @@ lazy val buildSettings = Seq(
         <email>dev-role@allenai.org</email>
       </developer>
     </developers>)
-) ++ PublishTo.sonatype
+) ++ releaseSettings ++
+  bintray.Plugin.bintrayPublishSettings ++
+  org.allenai.plugins.ReleasePlugin.SemanticVersion.settings
```
jkinkead: `ReleasePlugin` ought to be in-scope (you shouldn't need to qualify it).

publish.sbt
```diff
@@ -0,0 +1,35 @@
+import sbtrelease.ReleaseStep
+import sbtrelease.ReleaseStateTransformations._
+import sbtrelease.ReleasePlugin.ReleaseKeys
+
+
+bintray.Keys.repository in bintray.Keys.bintray in ThisBuild := "common"
+
+bintray.Keys.bintrayOrganization in bintray.Keys.bintray in ThisBuild := Some("allenai")
+
+val checkBranchIsNotMaster = { st: State =>
```
jkinkead: Would this be the code that eventually resides in a plugin?

jkinkead:
A few comments. I think the git revision specification might do what we want for merges (most?) all of the time, but the suggested specification looks more correct to me.

We should talk this over a bit. There's [a diagram on the git help page](http://schacon.github.io/git/git-rev-parse#_specifying_revisions) that really just convinced me I have no idea what `HEAD^{N > 1}` means.

jkinkead:
There's still one comment on the `build.sbt` file, not sure if you saw it.

Also, the question on `publish.sbt` - is that code that would maybe live in a plugin in the future?

Anyway, LGTM - merge after addressing the `build.sbt` comment (or not).

schmmd:
`publish.sbt` would move into `sbt-plugins`.

PR #112, opened 2015-04-30
Switch publishing to BinTray.  Make release 1.0.0.

schmmd:


README.md
```diff
@@ -14,39 +14,15 @@ A collection of useful utility classes and functions.
...
- 7. You should see the new version appear under https://oss.sonatype.org/content/repositories/releases/org/allenai/common/
+This project releases to BinTray.  To make a release, follow our standard
+release process.
+
+1.  Make sure you are on a branch other than `master`.
+2.  Set the upstream repository (`git push --set-upstream origin bintray`).
+3.  Kick of the release with `sbt release`.
+4.  Create a pull request and make sure Shippable builds it OK.
+5.  When the pull request is merged, Shippable will kick off a build and ultimately publish the artifact.
+6.  You can verify that it was published by looking on BinTray.com!
```
jkinkead: Feature request: Email after publish?

admin/publish-release.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
jkinkead: We should decide on the "official" version of these files, and include a link to it in the header. Once this is deployed and working outside of common, of course.

build.sbt
```diff
@@ -19,7 +18,7 @@ lazy val buildSettings = Seq(
         <email>dev-role@allenai.org</email>
       </developer>
     </developers>)
-) ++ PublishTo.sonatype
+) ++ bintray.Plugin.bintrayPublishSettings
```
jkinkead: Shouldn't this be provided by the new AI2 plugin?

jkinkead:
One real comment about maybe double-applying the bintray settings.

LGTM!

build.sbt
schmmd: I removed this from the bintray plugin as I said I might.  It will be annoying for projects that don't publish to bintray.

admin/publish-release.sh
schmmd: These are them ;-)

Good point--when they are rolled out elsewhere, add a comment that common contains the official versions?

schmmd:
I probably won't jump on that feature request, but it'd be great to have.

PR #113, opened 2015-04-30
Release 1.0.0

schmmd:


jkinkead:
LGTM.

Is there a tag on one of these commits? It seems like they're missing from the review UI.

PR #115, opened 2015-05-04
Add methods to safetly handle resources.

schmmd:


schmmd:
This was my idea for safe resource methods re: our recent conversation.

core/src/main/scala/org/allenai/common/Resource.scala
```diff
@@ -39,4 +41,47 @@ object Resource {
...
+    * @param  name  absolute path of the resource
+    */
+  def get(name: String): URL = {
+    getOpt(name).getOrElse {
+      throw new IllegalArgumentException("No such absolute resource found: " + name)
+    }
+  }
+
+  /** Get a Java Resource.
+    * This method provides a much nicer exception than the Java default (NPE).
```
jkinkead: This doesn't throw anything.

core/src/main/scala/org/allenai/common/Resource.scala
```diff
@@ -39,4 +41,47 @@ object Resource {
...
+    *
+    * @param  clazz  the class from which the resource may be relative
+    * @returns  Some(URL) if the resource exists
+    */
+  def getOpt(clazz: Class[_], name: String): Option[URL] = {
```
jkinkead: I think the more accepted scala pattern is to do like:
```scala
import scala.reflect.{ classTag, ClassTag }
def get[T : ClassTag](name: String): URL = {
  Option(classTag[T].runtimeClass.getResource(name))
}
```

core/src/main/scala/org/allenai/common/Resource.scala
```diff
@@ -39,4 +41,47 @@ object Resource {
...
+    getOpt(name).getOrElse {
+      throw new IllegalArgumentException("No such absolute resource found: " + name)
+    }
+  }
+
+  /** Get a Java Resource.
+    * This method provides a much nicer exception than the Java default (NPE).
+    *
+    * @param  name  absolute path of the resource
+    * @returns  Some(URL) if the resource exists
```
jkinkead: `@return`, not `@returns`. (. . . github tells me with its highlighting)

jkinkead:
Awesome.

A couple random comments, LGTM!

PR #117, opened 2015-05-14
added SingleInstanceUnitTest

cristipp:


cristipp:
@markschaake

markschaake:
:+1:

testkit/src/main/scala/org/allenai/common/testkit/SingeInstanceUnitSpec.scala
```diff
@@ -0,0 +1,6 @@
+package org.allenai.common.testkit
+
+/** Base class for any unit test that, for some reason, want to share the Spec instance.
+  * Usually the reason is a slow expensive test setup time (setting up a 3rd party DB, etc.)
+  */
+abstract class SingeInstanceUnitSpec extends AllenAiBaseSpec
```
cristipp: @markschaake @jkinkead Does this start to sound more like an integration test? It's cool we can reuse Scala test infrastructure for both unit and it, I just wonder whether we want to make the distinction more clear via the name of the base classes as well.

Something like IntegrationSpec. <end brainstorm>

markschaake:
Thanks @cristipp ! Merge on successful build?

markschaake:
oops, realized @jkinkead is the assignee

cristipp:
You both guys are awesome :)

testkit/src/main/scala/org/allenai/common/testkit/SingeInstanceUnitSpec.scala
markschaake: Would be good to have a `IntegrationSpec` too. I'd rather we kept it separate from the `SingleInstanceUnitSpec`. Could just be :

```scala
abstract class IntegrationSpec extends AllenAiBaseSpec
```

cristipp: Done.

jkinkead:
LGTM! Although you have a failed check of some sort.

cristipp:
Thanks for the feedback.

Is semaphore an alternative for shippable we are looking into? They seem to be broken, the error is "[error] (webapp/compile:compileIncremental) java.lang.NoSuchMethodError: scala.reflect.internal.StdNames$nme$.argument()Lscala/reflect/internal/Names$TermName;"

sbt test works for me.

markschaake:
> Is semaphore an alternative for shippable we are looking into?

@schmmd ?

schmmd:
I'm checking to see how well their caching works out of the box.

cristipp:
Ok, I'll go with Shippable green then :)

PR #121, opened 2015-06-02
Move ConfigModule helper class from solvers-common to AI2 common.

jkinkead:
This is a very useful helper class for using Typesafe Config, especially when consuming libraries.

I'd like to use this in the Wumpus client to make it easier to share configuration between Salience clients, so I'd like to have it here in common.


markschaake:
Since we don't use Guice everywhere, what do you think of putting this in a separate subproject and publish it as a separate artifact? Something like `common-guice` ?

jkinkead:
I feel mixed. I think we could get benefit from using this config module in lots of places, so putting it in mainline common is nice in that regard, and it's a small added dependency; but on the other hand, it *is* an added dependency that's not currently being used outside of Aristo.

Thoughts?

markschaake:
Yeah, it's the added dependency that pushes me in the separate subproject direction. I'm also imagining you might have some well-vetted Guice patterns that could belong in this other subproject as well and that it would be good to have a README dedicated to helping people with Guice for DI. Do you think you'd have more Guice-specific common code to add down the road?

jkinkead:
It's possible there'd be more code, but I can't think of anything at the moment. I never had the need to write custom code for Guice before - this module essentially replicates some of the features of Google's internal config system (based on commandline flags).

It would probably just end up being a subproject with a single source file, though. I'm fine with that, if you think it's the best way to go - it certainly makes for a good place to put documentation.

markschaake:
Sorry, thought I already replied to this. Let's make it a separate subproject for now, if you're ok with that.

jkinkead:
SGTM! I'll move it.

jkinkead:
Moved! TAL?

markschaake:
LGTM. Add a README to the subproject directory and merge?

jkinkead:
Done! Merging.

PR #123, opened 2015-06-05
Add a simple actor system module.

jkinkead:
Hah, I totally lied - there was at least once other Guice thing I wanted in common.


markschaake:
LGTM!

PR #126, opened 2015-06-12
Add resolver info to README.

schmmd:


cristipp:
Many thanks for the update!

PR #127, opened 2015-06-12
Update common plugin to pick up new dependencies.

jkinkead:
Care to review, Vu? This will let me publish an updated common-core, compiled against the latest spray-json.


vha14:
Can this be done in some (semi)automatic way in the future?

jkinkead:
Reliable automated dependency management requires a continuous integration system that can build *everything* at the same time to validate update. This isn't easy within the maven / sbt ecosystem.

vha14:
Got it. What should we do in the future to avoid this issue? 

I'll merge anyway since the issue seems non-trivial. 

Thanks Jesse!

jkinkead:
This particular issue is rooted in a bad version of `spray-json` - they released a minor version update that broken older clients. I don't think we can protect against this - if a library maintainer violates the semantic versioning contract, we're kind of screwed.

PR #128, opened 2015-06-12
Release v1.0.4

jkinkead:
This contains the updated spray dependencies, so that we can avoid some of the issues around the buggy version.


tusharkhot:
One character change!
LGTM

jkinkead:
The real review is for the git tag, which you can't see - it'll trigger a new release of the binaries. ;)

PR #129, opened 2015-06-12
Add the correct moduleName for publishing to bintray.

jkinkead:
I *think* this is what we need to make the post-publication merge step unnecessary.


jkinkead:
Sweet. That totally worked!

schmmd:
yay--my hero = Jesse

PR #131, opened 2015-06-18
Logging configuration

cristipp:
Thin wrapper to simplify logging configuration. Some defaults are already chosen, to point is to be easy to use, not to read 100 pages of docs to figure out which value goes into which slot. Open for future extension, once the additional needs become clearer.

In the big picture, the logback APIs are obscure, complex and verbose. logback.xml is uhm,  http://knowyourmeme.com/photos/552084-ha-ha-ha-no. Using programatic log configuration is unambiguous and can be easily customized for different scenarios.

APIs are not perfect, now it is the best moment to suggest improvements :)

Originating from https://github.com/allenai/euclid/blob/master/src/main/scala/org/allenai/euclid/Main.scala#L32

### DIFF
--- a/core/src/main/scala/org/allenai/common/Logging.scala
+++ b/core/src/main/scala/org/allenai/common/Logging.scala
@@ -61,8 +61,8 @@ trait Logging {
       * logger.Config("org.apache.spark").setLevel(Level.WARN)
       *
       * logger.Config().addAppender(
-      *   logger.Config.patternLayoutEncoder("%-5level [%thread]: %message%n"),
-      *   logger.Config.consoleAppender
+      *   logger.Config.newPatternLayoutEncoder("%-5level [%thread]: %message%n"),
+      *   logger.Config.newConsoleAppender
       * )
       * </code>
       *
@@ -91,11 +91,11 @@ trait Logging {
         * <code>
         * logger.Config()
         *   .addAppender(
-        *     logger.Config.patternLayoutEncoder("%-5level [%thread]: %message%n"),
-        *     logger.Config.consoleAppender)
+        *     logger.Config.newPatternLayoutEncoder("%-5level [%thread]: %message%n"),
+        *     logger.Config.newConsoleAppender)
         *   .addAppender(
-        *     logger.Config.htmlLayoutEncoder("%relative%thread%level%logger%msg"),
-        *     logger.Config.fileAppender("./log.html"))
+        *     logger.Config.newHtmlLayoutEncoder("%relative%thread%level%logger%msg"),
+        *     logger.Config.newFileAppender("./log.html"))
         * </code>
         */
       def addAppender(
@@ -115,13 +115,13 @@ trait Logging {
 
     /** Factory methods for some simple config objects. */
     object Config {
-      def patternLayoutEncoder(pattern: String): Encoder[ILoggingEvent] = {
+      def newPatternLayoutEncoder(pattern: String): Encoder[ILoggingEvent] = {
         val encoder = new PatternLayoutEncoder()
         encoder.setPattern(pattern)
         encoder
       }
 
-      def htmlLayoutEncoder(pattern: String): Encoder[ILoggingEvent] = {
+      def newHtmlLayoutEncoder(pattern: String): Encoder[ILoggingEvent] = {
         new LayoutWrappingEncoder[ILoggingEvent] {
           private val htmlLayout = new HTMLLayout()
           htmlLayout.setPattern(pattern)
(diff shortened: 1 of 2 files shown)


cristipp:
FYI @schmmd @gabrielStanovsky.

core/src/main/scala/org/allenai/common/Logging.scala
```diff
@@ -46,5 +52,107 @@ trait Logging {
...
+        encoder.start()
+        appender.setContext(loggerContext)
+        appender.setEncoder(encoder)
+        appender.start()
+        logger.addAppender(appender)
+      }
+    }
+
+    /** Factory methods for some simple config objects. */
+    object factory {
```
schmmd: I'd prefer `object Config`.  In general, use the same casing as Java recommends.  Factory methods are usually in the companion object of the class.

cristipp: Done. 
"factory" was feeling odd for me too. I was missing that I can use ```case class Config``` instead of ```def config()``` + ```class Config```.

core/src/main/scala/org/allenai/common/Logging.scala
```diff
@@ -46,5 +52,107 @@ trait Logging {
...
+      */
+    def config(loggerName: String = org.slf4j.Logger.ROOT_LOGGER_NAME) = {
+      new Config(loggerName)
+    }
+
+    class Config(loggerName: String = org.slf4j.Logger.ROOT_LOGGER_NAME) {
+      private val logger: Logger = LoggerFactory.getLogger(loggerName).asInstanceOf[Logger]
+
+      /** Resets the logger. */
+      def reset() = {
```
jkinkead: Return type, here and elsewhere?

core/src/main/scala/org/allenai/common/Logging.scala
```diff
@@ -46,5 +52,103 @@ trait Logging {
...
+        encoder.start()
+        appender.setContext(loggerContext)
+        appender.setEncoder(encoder)
+        appender.start()
+        logger.addAppender(appender)
+      }
+    }
+
+    /** Factory methods for some simple config objects. */
+    object Config {
```
jkinkead: Given that this symbol is going to be visible in any class extending `Logging`, and that we use typesafe's `Config` class all over the place, I'd prefer that this either be in the companion object to `Logging` or be renamed `LoggingConfig`.

core/src/main/scala/org/allenai/common/Logging.scala
```diff
@@ -46,5 +52,103 @@ trait Logging {
...
+      def patternLayoutEncoder(pattern: String) = {
```
jkinkead: Consider prefixing these with `new` or `create`.

jkinkead:
Some comments! Yay on the unit test! :D

core/src/main/scala/org/allenai/common/Logging.scala
cristipp: Done. Added a fluent interface, so its a bit easier to do multiple things to one Config.

core/src/main/scala/org/allenai/common/Logging.scala
cristipp: It's visible as "logger.Config", which is not that bad? Or at least in all the code I've seen, we ```extend Logging``` and use it as ```logger.info("my string")```. Now we're going also be able to do ```logger.Config(...).doSomething```. For example:

```
class Foo extends Logging { 
  logger.Config("org.apache").setLevel(Level.WARN)
}
```

Can you please show the proposed alternate syntax from client perspective?

cristipp: Perhaps:

```
class Foo extends Logging { 
  LoggingConfig("org.apache").setLevel(Level.WARN)
}
```

core/src/main/scala/org/allenai/common/Logging.scala
cristipp: Done.

cristipp:
Thanks for the comments. PTAL [Please Take Another Look]

core/src/main/scala/org/allenai/common/Logging.scala
jkinkead: Oh, you're right - I was confused here. I forgot this was in an internal `object`. `Config` LGTM!

jkinkead:
LGTM! I'll leave it up to you if you want to rename `Config` - but given that's it's scoped locally to `logger`, I'm fine with it as-is.

jkinkead:
The only comment I have is with the design - it's awkward to configure root logging from an arbitrary class, and that should be discouraged. Configuring local appenders / logging seems OK.

cristipp:
Thanks @jkinkead. I put everything under ```loggingConfig```, which is both discoverable (autocompletes when you type logger) and doesn't confuse package logger with global logger configuration.

jkinkead:
LGTM! Merging.

PR #136, opened 2015-06-23
Fix bintray name.

### LINKED ISSUE #135: bintray publishing is broken since 1.0.5
File from 1.0.4:  common-core_2.11-1.0.4-javadoc.jar 
File from 1.0.5 (broken): org.allenai.common:common-core_2.11_2.11-1.0.5-javadoc.jar 
File from 1.0.9 (broken): org.allenai.common:common-core_2.11_2.11-1.0.9-javadoc.jar

jkinkead:
This should be the correct setting to fix #135 .

The newest bintray plugin (`0.3.0`) has less terrible naming / design for these keys, so we should upgrade.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,8 @@
+import bintray.Keys.{ bintray => bintrayKey }
+
 lazy val buildSettings = Seq(
   organization := "org.allenai.common",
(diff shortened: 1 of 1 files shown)


jkinkead:
LGTM over hipchat . . .

PR #138, opened 2015-06-23
Override bintray name after applying bintray settings.

jkinkead:
This should make the bintray package name correct again.


cristipp:
LGTM. This all looks pretty chinese to me, but it feels good to see Jesse making progress :)

PR #144, opened 2015-07-06
added cacheKey: Option[String] to Version

ryanai3:
Adds cacheKey as an option field of version and handles building from resources

### DIFF
--- a/core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
+++ b/core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
@@ -4,9 +4,7 @@ import java.io.File
 import org.allenai.common.{ GitVersion, Version, cache }
 import org.allenai.common.testkit.UnitSpec
 import org.scalatest.BeforeAndAfterAll
-import spray.json.DefaultJsonProtocol
 import spray.json.DefaultJsonProtocol._
-import spray.json._
 import sys.process._
 
 case class Foo(stringVar: String, intVar: Int)
@@ -16,11 +14,13 @@ object FooJsonProtocol extends DefaultJsonProtocol {
 }
 
 class QueryCaches(redisHostname: String, redisPort: Int) {
+  import FooJsonProtocol.__
+
   val stringQueryCache = new JsonQueryCache[String](redisHostname, redisPort, "test")
   val intQueryCache = new JsonQueryCache[Int](redisHostname, redisPort, "test")
   val seqStringQueryCache = new JsonQueryCache[Seq[String]](redisHostname, redisPort, "test")
+
   // It's an object I can test
-  import FooJsonProtocol._
   val fooQueryCache = new JsonQueryCache[Foo](redisHostname, redisPort, "test")
 
   val stringKey = "stringKey"
@@ -88,14 +88,10 @@ class QueryCacheSpec extends UnitSpec with BeforeAndAfterAll {
   override def afterAll() {
     Seq("redis-cli", "FLUSHALL").!!
     Seq("redis-cli", "SHUTDOWN").!!
-    // delete redis dump file if it exists
-    try {
-      val f = new File("./dump.rdb")
-      if (f.exists) {
-        f.delete()
-      }
-    } catch {
-      case e: Throwable => ""
+    // delete redis dump file if it exists 
+    val f = new File("./dump.rdb")
+    if (f.exists) {
+      f.delete()
     }
   }
 }
--- a/core/src/main/scala/org/allenai/common/Version.scala
+++ b/core/src/main/scala/org/allenai/common/Version.scala
@@ -59,7 +59,7 @@ object GitVersion {
   *
   * @param  git  the git version (commit information) of the build.
   * @param  artifactVersion  the version of the artifact in the build.
-  * @param  cacheKey a cacheKey of the project. changes on git commits to src of project and dependency changes
+  * @param  cacheKey a cacheKey of the project. Changes on git commits to src of project and dependency changes.
   */
 case class Version(
     git: GitVersion,
--- a/core/src/main/scala/org/allenai/common/Version.scala
+++ b/core/src/main/scala/org/allenai/common/Version.scala
@@ -62,7 +62,8 @@ object GitVersion {
   */
 case class Version(
     git: GitVersion,
-    artifactVersion: String
+    artifactVersion: String,
+    cacheKey: Option[String] = None
 ) {
   @deprecated("Use artifactVersion instead.", "2014.09.09-1-SNAPSHOT")
   def artifact = artifactVersion
@@ -81,19 +82,26 @@ object Version {
     val pkg = "/" + org + "/" + name.replaceAll("-", "")
     val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
     val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-
+    val cacheKeyConfUrl = this.getClass.getResource(pkg + "/cacheKey.conf")
     require(artifactConfUrl != null, "Could not find artifact.conf in " + pkg + ".")
     require(gitConfUrl != null, "Could not find git.conf in " + pkg + ".")
 
     val artifactConf = ConfigFactory.parseURL(artifactConfUrl)
     val gitConf = ConfigFactory.parseURL(gitConfUrl)
-
     val artifactVersion = artifactConf[String]("version")
     val sha1 = gitConf[String]("sha1")
     val commitDate = gitConf[Long]("date")
     val remotes = gitConf.getStringList("remotes").asScala
 
-    Version(GitVersion.create(sha1, commitDate, remotes), artifactVersion)
+    val cacheKey = cacheKeyConfUrl match {
+      case null => None
+      case _ => {
+        val cacheKeyConf = ConfigFactory.parseURL(cacheKeyConfUrl)
+        Some(cacheKeyConf[String]("cacheKey"))
+      }
+    }
+
+    Version(GitVersion.create(sha1, commitDate, remotes), artifactVersion, cacheKey)
   }
 
   import spray.json.DefaultJsonProtocol._
--- a/core/src/main/scala/org/allenai/common/Version.scala
+++ b/core/src/main/scala/org/allenai/common/Version.scala
@@ -83,7 +83,7 @@ object Version {
     val pkg = "/" + org + "/" + name.replaceAll("-", "")
     val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
     val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-    val cacheKeyConfUrl = this.getClass.getResource(pkg + "/cacheKey.conf")
+    val cacheKeyConfUrl = this.getClass.getResource(pkg + "/cacheKey.Sha1")
     require(artifactConfUrl != null, "Could not find artifact.conf in " + pkg + ".")
     require(gitConfUrl != null, "Could not find git.conf in " + pkg + ".")
 
@@ -93,7 +93,9 @@ object Version {
     val sha1 = gitConf[String]("sha1")
     val commitDate = gitConf[Long]("date")
     val remotes = gitConf.getStringList("remotes").asScala
-    val cacheKey = Option(cacheKeyConfUrl) map { url => ConfigFactory.parseURL(url)[String]("cacheKey") }
+    val cacheKey = Option(cacheKeyConfUrl) map { url =>
+      scala.io.Source.fromURL(cacheKeyConfUrl).getLines().mkString.trim
+    }
     Version(GitVersion.create(sha1, commitDate, remotes), artifactVersion, cacheKey)
   }
 
--- a/core/src/main/scala/org/allenai/common/Version.scala
+++ b/core/src/main/scala/org/allenai/common/Version.scala
@@ -93,9 +93,7 @@ object Version {
     val sha1 = gitConf[String]("sha1")
     val commitDate = gitConf[Long]("date")
     val remotes = gitConf.getStringList("remotes").asScala
-    val cacheKey = Option(cacheKeyConfUrl) map { url =>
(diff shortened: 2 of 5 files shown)


ryanai3:
@jkinkead please review - need 2 reviewers

schmmd:
This code seems Aristo specific.  Does it really belong in common?  How is cacheKey.conf specified?

ryanai3:
current use-case is aristo specific, but cacheKey.conf is a hash for the current sbt-project, so it's a notion of version for non-solver project's too.

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -81,19 +82,26 @@ object Version {
...
     val artifactConf = ConfigFactory.parseURL(artifactConfUrl)
     val gitConf = ConfigFactory.parseURL(gitConfUrl)
-
     val artifactVersion = artifactConf[String]("version")
     val sha1 = gitConf[String]("sha1")
     val commitDate = gitConf[Long]("date")
     val remotes = gitConf.getStringList("remotes").asScala
 
-    Version(GitVersion.create(sha1, commitDate, remotes), artifactVersion)
+    val cacheKey = cacheKeyConfUrl match {
```
jkinkead: You can use `Option.apply` here instead:
```scala
val cacheKey = Option(cacheKeyConfUrl) map { url => ConfigFactory.parseURL(url)[String]("cacheKey") }
```

jkinkead:
`cacheKey.conf` will be set in the release plugin, at least for the time being. It depends on the `stage` command.

This is logically a version hash (hash of deployed code + dependencies), so having it in `Version` makes some sense - it could be useful for any code that wants to use a cache.

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -62,7 +62,8 @@ object GitVersion {
   */
 case class Version(
     git: GitVersion,
-    artifactVersion: String
+    artifactVersion: String,
+    cacheKey: Option[String] = None
```
jkinkead: Add documentation for this. Also, I'd avoid having a default here - `Version` should be created through one of the local helpers.

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -59,10 +59,12 @@ object GitVersion {
   *
   * @param  git  the git version (commit information) of the build.
   * @param  artifactVersion  the version of the artifact in the build.
+  * @param  cacheKey a cacheKey of the project. changes on git commits to src of project and dependency changes
```
jkinkead: Scaladoc follows javadoc style conventions - this should have the second sentence capitalized and ending with a period - `Changes ... and dependency changes.`

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -81,19 +83,18 @@ object Version {
     val pkg = "/" + org + "/" + name.replaceAll("-", "")
     val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
     val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-
+    val cacheKeyConfUrl = this.getClass.getResource(pkg + "/cacheKey.conf")
```
jkinkead: Hrm, I realized a problem with this: You're generating this as a resource file, but you're also generating it *after* you've already built the jar, so it won't get embedded.

You should be generating the cache key file within the staging directory, after the build is staged, and reading it from a system property path.

Right now, the deploy task looks something like:

  * Run in parallel:
    * stage -> generate cache key
    * clean
    * stage

which is not good. There should be only one stage step.

You should update the deploy plugin to have a task that replaces [the stage dependency](https://github.com/allenai/sbt-plugins/blob/master/src/main/scala/org/allenai/plugins/DeployPlugin.scala#L191) that does something like:
```scala
newTask := {
  val stageDir = (UniversalPlugin.autoImport.stage in thisProject).value
  // Checksum stage dir, gen cache file in `conf` or `lib` or something.
  stageDir
}
```

ryanai3: Doesn't the deployplugin have 2 stage steps even without cacheKey generation?
// Clean up anything leftover in the staging directory before re-staging.
    UniversalPlugin.autoImport.stage <<=
      UniversalPlugin.autoImport.stage.dependsOn(cleanStage),

jkinkead: It shouldn't - that should just make it remove the old staging dir before starting the main stage.

ryanai3:
@jkinkiead, do you have any examples of code reading in files from the rsynced conf directory or any of the rsynced directories? I've been looking through the repo but can't find much

ryanai3:
@jkinkead pls review

jkinkead:
You've got build errors still . . .

ryanai3:
@jkinkead recent commit history is ugly... should I squash?

jkinkead:
What squashing does in a PR is hide any comments made to files that don't exist in the commit history any more. So, if you're squashing up to 4f28f8a (the info route commit), that's pretty harmless.

jkinkead:
This looks like it needs a couple of updates? The cache key file isn't named correctly (it's called `.conf` instead of `.Sha1`). There also might be a problem with how it's loading the cache file?

ryanai3:
cache key file isn't named correctly, you're right. Is using a configfactory not the proper way to load it?
that's the way it's done for artifactConf & gitConf?
It is just a line of text, I could just read it in

jkinkead:
Yeah, I was thinking just looking for the `.sha1` file, and loading it if it's around. It can just be an opaque string - it doesn't need to have a key / value you read out.

ryanai3:
@jkinkead look over this one!

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -81,19 +83,18 @@ object Version {
     val pkg = "/" + org + "/" + name.replaceAll("-", "")
     val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
     val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-
+    val cacheKeyConfUrl = this.getClass.getResource(pkg + "/cacheKey.Sha1")
```
jkinkead: No longer used, should be removed.

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -81,19 +83,18 @@ object Version {
...
     val artifactConf = ConfigFactory.parseURL(artifactConfUrl)
     val gitConf = ConfigFactory.parseURL(gitConfUrl)
-
     val artifactVersion = artifactConf[String]("version")
     val sha1 = gitConf[String]("sha1")
     val commitDate = gitConf[Long]("date")
     val remotes = gitConf.getStringList("remotes").asScala
-
-    Version(GitVersion.create(sha1, commitDate, remotes), artifactVersion)
+    val cacheKey = Option(System.getProperty("application.cacheKey", null))
```
jkinkead: I think `null` is the default anyway?

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -1,17 +1,27 @@
 package org.allenai.common.cache
 
+import java.io.File
 import org.allenai.common.{ GitVersion, Version, cache }
 import org.allenai.common.testkit.UnitSpec
 import org.scalatest.BeforeAndAfterAll
-import sys.process._
+import spray.json.DefaultJsonProtocol
 import spray.json.DefaultJsonProtocol._
```
jkinkead: You shouldn't need this import anymore, since you're now importing from the `FooJsonProtocol`.

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -1,17 +1,27 @@
...
+import spray.json._
```
jkinkead: This just gives you `toJson` and `parseJson`. I don't think you're using those?

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -78,5 +88,14 @@ class QueryCacheSpec extends UnitSpec with BeforeAndAfterAll {
   override def afterAll() {
     Seq("redis-cli", "FLUSHALL").!!
     Seq("redis-cli", "SHUTDOWN").!!
+    // delete redis dump file if it exists
+    try {
```
jkinkead: I wouldn't do a try-catch here - this should be safe.

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -1,17 +1,27 @@
...
 
+import java.io.File
 import org.allenai.common.{ GitVersion, Version, cache }
 import org.allenai.common.testkit.UnitSpec
 import org.scalatest.BeforeAndAfterAll
-import sys.process._
+import spray.json.DefaultJsonProtocol
 import spray.json.DefaultJsonProtocol._
+import spray.json._
+import sys.process._
```
jkinkead: Unused.

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -1,17 +1,27 @@
...
+}
 
 class QueryCaches(redisHostname: String, redisPort: Int) {
   val stringQueryCache = new JsonQueryCache[String](redisHostname, redisPort, "test")
   val intQueryCache = new JsonQueryCache[Int](redisHostname, redisPort, "test")
   val seqStringQueryCache = new JsonQueryCache[Seq[String]](redisHostname, redisPort, "test")
   // It's an object I can test
-  val versionQueryCache = new JsonQueryCache[Version](redisHostname, redisPort, "test")
+  import FooJsonProtocol._
+  val fooQueryCache = new JsonQueryCache[Foo](redisHostname, redisPort, "test")
```
jkinkead: Does this talk over the network? If so, it should be under `it` (and maybe not run automatically).

webapp/src/main/scala/org/allenai/common/webapp/InfoRoute.scala
```diff
@@ -25,7 +25,8 @@ class InfoRoute(val info: Map[String, String] = Map.empty) {
           "gitDatePretty" -> version.git.prettyCommitDate
         ) ++
           version.git.repoUrl.map("gitRepoUrl" -> _) ++
-          version.git.commitUrl.map("gitCommitUrl" -> _)
+          version.git.commitUrl.map("gitCommitUrl" -> _) ++
+          Option(System.getProperty("application.cacheKey")).map { value => Map("cacheKey" -> value) }.getOrElse(Map())
```
jkinkead: Why aren't you doing the same thing as above (iterating over an `Option`)?
```scala
Option(System.getProperty("application.cacheKey")).map("cacheKey" -> _)
```

jkinkead: Also, this should use the cache key from the provided `Version`.

jkinkead:
A bunch of small style comments.

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: yup, moved

core/src/main/scala/org/allenai/common/Version.scala
ryanai3: done

core/src/main/scala/org/allenai/common/Version.scala
ryanai3: it is - fixed

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: done

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: fixed

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: done

core/src/test/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: sys.process is needed to shutdown redis - it's the 
```scala
Seq("redis-cli", "FLUSHALL").!!
Seq("redis-cli", "SHUTDOWN").!!
``` calls

**Comment by ryanai3** [comment]
@jkinkead fixes!

File: core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala [hunk 13]
```diff
@@ -1,17 +1,26 @@
...
+  implicit val fooFormat = jsonFormat2(Foo)
+}
 
 class QueryCaches(redisHostname: String, redisPort: Int) {
   val stringQueryCache = new JsonQueryCache[String](redisHostname, redisPort, "test")
   val intQueryCache = new JsonQueryCache[Int](redisHostname, redisPort, "test")
   val seqStringQueryCache = new JsonQueryCache[Seq[String]](redisHostname, redisPort, "test")
   // It's an object I can test
-  val versionQueryCache = new JsonQueryCache[Version](redisHostname, redisPort, "test")
+  import FooJsonProtocol._
```
jkinkead: If you move this import to the top of `QueryCaches`, you won't need the `DefaultJsonProtocol._` import.

jkinkead:
One optional comment, merge when you're happy!

core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
ryanai3: didn't see that! you're right

PR #151, opened 2015-09-15
New SeekableSource class for efficient file reading.

jkinkead:
On my laptop, this outperforms `scala.io.Source` reading character-wise by a factor of 3.4, and line-wise by a factor of 1.7.

If you want to see the slightly out-of-date version with a memory map option, you can check out [this commit](https://github.com/jkinkead/common/tree/2006bb1c2ee0851ef9c978a52347ff43a670cab5). That's a little slower for the line iterator, because it's not using the dirty tricks I gleaned from BufferedReader. It's still valid for comparing the memory map option with the standard option, though.

### DIFF
--- a/core/src/main/scala/org/allenai/common/SeekableSource.scala
+++ b/core/src/main/scala/org/allenai/common/SeekableSource.scala
@@ -12,16 +12,20 @@ import scala.io.Codec
   * either UTF-8 or ISO-8859-1 encoded.
   * <p>This also has a `getLines()` method to iterate over lines. Any reads done on this iterator
   * are reflected in the main Source, and reads on the Source are reflected in the
-  * `Iterator[String]`.</p>
+  * `Iterator[String]`. The iterator returned by `getLines()` will create another buffer of
+  * `bufferSize` bytes, so bear this in mind when estimating memory usage.</p>
   * @param inFile the file channel to wrap
-  * @param bufferSize the size of the internal buffer to use
+  * @param bufferSize the size of the internal buffer to use. Defaults to 8MB.
   * @param codec the codec to use. Must be one of UTF-8 or ISO-8859-1.
-  * @throws IllegalArgumentException if `bufferSize` is less than 3, or `codec` is not UTF-8 or
+  * @throws IllegalArgumentException if `bufferSize` is less than 4, or `codec` is not UTF-8 or
   * ISO-8859-1
   */
-class SeekableSource(inFile: FileChannel, bufferSize: Int = 8192)(implicit codec: Codec)
+class SeekableSource(inFile: FileChannel, bufferSize: Int = 8 << 20)(implicit codec: Codec)
     extends Iterator[Char] {
-  require(bufferSize >= 3, "Buffer must be at least 3 bytes to decode UTF-8!")
+  require(bufferSize >= 4, "Buffer must be at least 4 bytes to decode UTF-8!")
+
+  // "Unknown" character. Used as a replacement for bad characters encountered.
+  val BadChar = 0xfffd
 
   // The motivation for this class is explained above, but some of the implementation choices are a
   // little confusing. This has decoding implemented within the class below, which seems odd. The
@@ -51,6 +55,9 @@ class SeekableSource(inFile: FileChannel, bufferSize: Int = 8192)(implicit codec
     buffer.limit(0)
     buffer
   }
+
+  /** True if we're in the middle of a two-Java-char unicode character. */
+  private[common] var wideCharBytesRemaining = false
 
   /** Returns true if the source has more input, populating the input buffer if need be. */
   override def hasNext: Boolean = inBuffer.hasRemaining || (inputRemaining && fillBuffer())
@@ -91,7 +98,7 @@ class SeekableSource(inFile: FileChannel, bufferSize: Int = 8192)(implicit codec
   /** @return the next character in the buffer, decoded from UTF-8 */
   protected[common] def nextUtf8: Char = {
     // Ensure we are either at the end of file, or have at least three bytes that are readable.
-    if (inBuffer.remaining() < 3 && inputRemaining) {
+    if (inBuffer.remaining() < 4 && inputRemaining) {
       fillBuffer()
     }
     // See https://en.wikipedia.org/wiki/UTF-8 for details. This is checking the first byte for the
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/allenai/common/SeekableSource.scala
```diff
@@ -0,0 +1,256 @@
...
+  * <p>This also has a `getLines()` method to iterate over lines. Any reads done on this iterator
+  * are reflected in the main Source, and reads on the Source are reflected in the
+  * `Iterator[String]`.</p>
+  * @param inFile the file channel to wrap
+  * @param bufferSize the size of the internal buffer to use
+  * @param codec the codec to use. Must be one of UTF-8 or ISO-8859-1.
+  * @throws IllegalArgumentException if `bufferSize` is less than 3, or `codec` is not UTF-8 or
+  * ISO-8859-1
+  */
+class SeekableSource(inFile: FileChannel, bufferSize: Int = 8192)(implicit codec: Codec)
```
dirkgr: Couldn't you have a buffer much larger than this? Why not 8MB?

core/src/main/scala/org/allenai/common/SeekableSource.scala
```diff
@@ -0,0 +1,256 @@
...
+    // Ensure we are either at the end of file, or have at least three bytes that are readable.
+    if (inBuffer.remaining() < 3 && inputRemaining) {
+      fillBuffer()
+    }
+    // See https://en.wikipedia.org/wiki/UTF-8 for details. This is checking the first byte for the
+    // total length of the encoding.
+    val first = inBuffer.get()
+    val intVal = if ((first & 0x80) == 0) {
+      return first.toChar
+    } else if ((first & 0xe0) == 0xc0) {
```
dirkgr: This is conditional of the byte starting with `110`. Is that what you want? It's not what the comment says below.

core/src/main/scala/org/allenai/common/SeekableSource.scala
```diff
@@ -0,0 +1,256 @@
...
+      // TODO(jkinkead): Secondary bytes (here and below) should all have their two highest bits set
+      // as 10. This should be asserted somehow, and if they're corrupted, producing an unknown
+      // character instead.
+      val second = inBuffer.get()
+      ((first & 0x1f) << 6) | (second & 0x3f)
+    } else if ((first & 0xf0) == 0xe0) {
```
dirkgr: And this starts with `1110`. Maybe say that somewhere?

core/src/main/scala/org/allenai/common/SeekableSource.scala
```diff
@@ -0,0 +1,256 @@
...
+      // won't fit in a JVM Char.
+
+      // skipCount stores the *additional* bytes we will skip, not the total length of the encoding.
+      val skipCount = if ((first & 0xf8) == 0xf0) {
+        3
+      } else if ((first & 0xfc) == 0xf8) {
+        4
+      } else if ((first & 0xfe) == 0xfc) {
+        5
+      } else {
```
dirkgr: I read somewhere that UTF-8 was designed so you can pick up a stream in the middle. There is some way to figure out where the next character starts. If that's true, you shouldn't have to do this. On the other hand, we're in the realm of illegal UTF-8, so maybe that doesn't apply here anyways.

core/src/test/scala/org/allenai/common/SeekableSourceSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+import java.nio.file.{ Files, StandardOpenOption }
+
+import scala.io.Codec
+
+class SeekableSourceSpec extends UnitSpec {
+  // UTF-8 is the default encoding for these tests.
+  implicit val codec = Codec.UTF8
+
+  /** Stores füü, in UTF-8. */
+  val foomlaut: Array[Byte] = Array('f'.toByte, 0xc3, 0xbc, 0xc3, 0xbc) map { _.toByte }
```
dirkgr: What about characters that span three bytes?

core/src/test/scala/org/allenai/common/SeekableSourceSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+    val windowsNewlines = newFileWithChars("foo\rbar\r\ngaz\r")
+    val source = new SeekableSource(windowsNewlines)
+    val lines = source.getLines()
+
+    lines.next() should be("foo")
+    lines.next() should be("bar")
+    lines.next() should be("gaz")
+    lines.hasNext should be(false)
+  }
+}
```
dirkgr: What about a test for the malformed UTF-8 case?

dirkgr:
I'm assuming you smoke-tested this on the Wikipedia corpus, to make sure it doesn't blow up?

jkinkead:
Yeah, my test numbers come from reading the wikipedia corpus - and verifying line counts with `scala.io.Source`.

core/src/main/scala/org/allenai/common/SeekableSource.scala
jkinkead: I think I'm just living in the past . . . :P

This size is what `java.io.BufferedReader` uses. Otherwise, it's pretty arbitrary. I wasn't seeing huge boosts when I was playing around with sizes, but 8M seems like a totally sensible default.

core/src/main/scala/org/allenai/common/SeekableSource.scala
jkinkead: The comment is talking about secondary bytes (all bytes following the first size-indicating byte). I'll try to clarify (and address the TODO).

core/src/main/scala/org/allenai/common/SeekableSource.scala
jkinkead: Done.

core/src/main/scala/org/allenai/common/SeekableSource.scala
jkinkead: Actually, yes, you can skip any bytes that start with the secondary-character sigil. I'll do that.

core/src/test/scala/org/allenai/common/SeekableSourceSpec.scala
jkinkead: Done.

core/src/test/scala/org/allenai/common/SeekableSourceSpec.scala
jkinkead: Done.

jkinkead:
OK, comments addressed, and unit tests expanded! I also fixed a couple of bugs.

It turns out that Java *does* support all characters in UTF-16, so I updated the by-character reader to accommodate this. Did you want to take another look?

dirkgr:
UTF is messy business.

dirkgr:
When you have all the special cases handled, make sure it's still faster than what you're replacing :-)

jkinkead:
What's really strange is that the comments you made on the specific commits went to my gmail account . . . :)

I'll let you know when I've added tests for the other corner-cases (and fixed them if need be).

jkinkead:
Special-case tested (and handled).

dirkgr:
At this point, I have no further ideas about how to break this. LGTM!

jkinkead:
Sweet!

PR #154, opened 2015-10-05
Add scaladoc publication settings, and clean up scaladoc

jkinkead:
Pretty straightforward! I think we'll want this automatically run as part of our release process.

Scaladoc published [here](http://allenai.github.io/common/latest/api/).


dirkgr:
LGTM.

We do this with all our APIs, and they publish to the same place? How do we keep the versions straight if that's what we do?

The title of the API docs is "root - root", which isn't very informative.

jkinkead:
They publish to project-specific places. In theory, we could publish a version with each release with a specific URL, but I didn't want to sink any more time into SBT nonsense than I had to. I'll definitely look into it when I make this part of the release process.

I'm not sure how to change the title without adding documentation for the package `org`, which seems icky.

PR #159, opened 2015-11-18
Make Enum java serializable

markschaake:
Our `Enum` was not serializable. This caused runtime exceptions during Spark pipeline runs due to failed java serialization during closure compilation.

@vha14 I'm pretty sure this will fix any java serialization issues we have with our Enum's. The unit test I added failed before having `Enum` extend `Serializable`

@dirkgr FYI

### DIFF
--- a/core/src/test/scala/org/allenai/common/EnumSpec.scala
+++ b/core/src/test/scala/org/allenai/common/EnumSpec.scala
@@ -4,15 +4,13 @@ import org.allenai.common.testkit.UnitSpec
 
 import spray.json._
 
-class EnumSpec extends UnitSpec {
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.io.ObjectInputStream
+import java.io.ObjectOutputStream
+import java.nio.file.Files
 
-  sealed abstract class FakeEnum(name: String) extends Enum[FakeEnum](name)
-  object FakeEnum extends EnumCompanion[FakeEnum] {
-    case object Value1 extends FakeEnum("value1")
-    case object Value2 extends FakeEnum("value2")
-    case object Value3 extends FakeEnum("value3")
-    register(Value1, Value2, Value3)
-  }
+class EnumSpec extends UnitSpec {
 
   "all" should "return all registerd Enum's" in {
     assert(FakeEnum.all.size === 3)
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/allenai/common/EnumSpec.scala
```diff
@@ -4,15 +4,11 @@ import org.allenai.common.testkit.UnitSpec
 
 import spray.json._
 
-class EnumSpec extends UnitSpec {
+import java.io.{ FileOutputStream, FileInputStream }
```
jkinkead: Put all `java.io` imports in one import block, and sort them?

markschaake: they are in the import block. Are you misreading the diff?

cristipp: Drive by style comment: I rely on Intellij to manage imports. The only
thing I do manually is to remove unused imports, conveniently marked in
grey. In XX years of Java / Scala, I never cared once about the order of
imports; I blank them out harder than ads on the Web. I get it's nice to
have tidy imports, but barring a good autoformat, should we spend
nano-attention bits on this?

markschaake: Sticking to our import formatting standards I think is a good policy for code readability (especially code review and approaching code that you haven't seen before). There is a real cost without simple automated tooling (agreeing with Cristian - evidence is the cost of having this conversation which is one that has been had many times at AI2). I think we should either finally invest in automating import formatting or decide to drop the import formatting standard. My vote would be for investing in automation. We already have Jesse's git commit hook scalariform formatter. Would be great if we could hook into that (no pun intended) with an auto import formatter. Not sure how hard it would be to write a tool that did this. I've written some [elisp functions](https://github.com/markschaake/emacs-conf/blob/master/ai2-imports.el) that do this for me Emacs. Could be ported to scala easily and implemented as an SBT plugin (for example).

cristipp: https://github.com/daniel-trinh/scalariform/issues/66

jkinkead: Autoimporting is not possible in scalariform due to the fact that the order of import statements matters - you can import from local symbols - and scalariform is not a full compiler. However, it would be easy to add a script which did this heuristically, and would be guaranteed correct without relative imports.

@markschaake - I was referring to the two `import java.io.{ }` blocks. They should be merged, yes?

markschaake: I see, I thought you thought the imports were inside the class declaration. I'll fix.

markschaake:
@jkinkead I fixed the imports. I just converted to non-block imports and sorted.

PR #161, opened 2015-11-19
Relax self-type of ScratchDirectory

markschaake:
@jkinkead mind taking this one?

Motivation: we have a `UnitSpec` that requires one-instance-per-test. I want to use the `ScratchDirectory` in a test that does not create a new instance per test. Relaxing the self type allows me to do that.

Also, fixed integration test compile error. Should probably add `sbt it:compile` to the Semaphore build to prevent this (I'll do this after merge).

### DIFF
--- a/core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
+++ b/core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
@@ -4,7 +4,7 @@ import java.io.File
 import org.allenai.common.{ GitVersion, Version, cache }
 import org.allenai.common.testkit.UnitSpec
 import org.scalatest.BeforeAndAfterAll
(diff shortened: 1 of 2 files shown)


core/src/it/scala/org/allenai/common/cache/QueryCacheSpec.scala
```diff
@@ -14,7 +14,7 @@ object FooJsonProtocol extends DefaultJsonProtocol {
 }
 
 class QueryCaches(redisHostname: String, redisPort: Int) {
-  import FooJsonProtocol.__
+  import FooJsonProtocol._
```
markschaake: Broken window fix here: this file wasn't compiling

jkinkead:
LGTM! Merging.

PR #164, opened 2015-12-01
Take in optional Overrides to Index-builder Config parameters

sbhaktha:
@row11 : here it is.

### DIFF
--- a/indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
+++ b/indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
@@ -32,7 +32,13 @@ import java.util.concurrent.TimeUnit
   * Refer http://joelabrahamsson.com/elasticsearch-101/ to get started.
   * Takes in Config object containing corpus and other information necessary to build the index.
   */
-class BuildCorpusIndex(config: Config) extends Logging {
+class BuildCorpusIndex(originalConfig: Config, configOverrides: Option[Config]) extends Logging {
+
+  /** Get merged Config object from applying requested overrides to original config. */
+  val config = configOverrides match {
+    case Some(overrides) => overrides.withFallback(originalConfig)
+    case None => originalConfig
+  }
 
   /** Get Index Name and Index Type. */
   val esConfig: Config = config[Config]("elasticSearch")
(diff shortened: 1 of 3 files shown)


indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndexRunner.scala
```diff
@@ -1,27 +1,33 @@
...
       c.copy(indexConfigName = x)
     } text ("name of configuration to use")
+    opt[File]("config-overrides-file") valueName ("<file>") action { (x, c) =>
+      c.copy(configOverrideFile = Some(x))
+    } text ("Config file with overrides if any")
   }
 
   parser.parse(args, IndexConfig()).foreach(config => {
     val rootConfig = ConfigFactory.parseResources(getClass, "indexing.conf").resolve()
+    val configOverrides = config.configOverrideFile map { f => ConfigFactory.parseFile(f) }
```
row11: I would personally put the pattern matching here that way you don't have to change the signature for `BuildCorpusIndex`. I mean something like:
```
val originalConfig = rootConfig.getConfig(s"org.allenai.common.indexing.${config.indexConfigName}")
val finalConfig = configOverrides match {
    case Some(overrides) => overrides.withFallback(originalConfig)
    case None => originalConfig
}
new BuildCorpusIndex(finalConfig.resolve()).buildElasticSearchIndex()
```
But up to you!

row11:
One suggestion, otherwise LGTM!

indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
```diff
@@ -32,7 +32,13 @@ import java.util.concurrent.TimeUnit
   * Refer http://joelabrahamsson.com/elasticsearch-101/ to get started.
   * Takes in Config object containing corpus and other information necessary to build the index.
   */
-class BuildCorpusIndex(config: Config) extends Logging {
+class BuildCorpusIndex(originalConfig: Config, configOverrides: Option[Config]) extends Logging {
```
row11: if you do decide to leave this, maybe have `configOverrides` default to `None`?

sbhaktha:
Good idea. Done!

sbhaktha:
And Merging since it LGTY! :)

PR #165, opened 2015-12-08
Add segmenting by paragraph as an option to BuildCorpusIndex

matt-gardner:
Also, added a node about running elasticsearch in the README, and updated the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .cache
+.cache-main
+*.swp
 .classpath
 .project
 .settings
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 .cache
-.cache-main
-*.swp
 .classpath
 .project
 .settings
--- a/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
+++ b/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
@@ -61,7 +61,6 @@ org.allenai.common.indexing.paragraph_base: ${org.allenai.common.indexing.base} 
           text: {
             type: "multi_field",
             fields: {
-              // Not sure yet what this "analyzer" field means.
               text: {type: "string", analyzer: "snowball"}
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
 .cache
+.cache-main
+*.swp
```
jkinkead: It's better practice to put these kinds of things in `~/.gitignore` so they don't need to be in every project.

I also recomment `*.sw[pon]`, for those cases where you have multiple edits. :)

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -39,16 +49,35 @@ org.allenai.common.indexing.base: {
...
+  elasticSearch: {
+    indexType: "paragraph"
+    mapping: {
+      paragraph: {
+        dynamic: false,
+        properties: {
+          text: {
+            type: "multi_field",
+            fields: {
+              // Not sure yet what this "analyzer" field means.
```
kwhumphreys: specifies how to preprocess text before indexing. the snowball analyzer does case normalization, stopword removal, and stemming.

matt-gardner: I used a browser to search my local elasticsearch index (https://github.com/mobz/elasticsearch-head), and using capital letters made all the results disappear.  This would be why?

kwhumphreys: if your browser searches the 'text' field rather than the 'text_raw' field, yes

matt-gardner: Thanks, that makes more sense now.  Comment removed.

matt-gardner:
After I looked at other corpora, I realized this was more complicated than I originally thought.  It needs a more general solution.  I'm closing this PR for now.

PR #166, opened 2015-12-10
Allow for segmenting by paragraph in BuildCorpusIndex

matt-gardner:
The biggest change here is in BuildCorpusIndex.scala.  There used to be three paths through the code for three different corpus specifications, with a lot of code duplicated between those three paths.  I merged them into a single code path, where first you determine what kind of a corpus to index (local vs. from the datastore, what the document format is), then you process the file according to its document format.

After this refactoring, I added some new document formats that permit segmenting by paragraph instead of by sentence.  I have run this locally for the Barron's study guide and for simple wikipedia (both by sentence, to make sure that still works, and the new paragraph indexing), and it works for me.  I didn't run it on the waterloo corpus or full wikipedia, because I thought it was too big for my laptop to handle, and I don't know how to use EC2 for this.

Part of the new document format stuff involves a new document representation, SegmentedDocument.  This isn't a perfect document representation, but it's good enough for what I needed.  I'm not sure if this is the right place to put this class, but it does seem rather specific to the indexing code.

I also added tests for most of the code that I touched.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .cache
+.cache-main
+*.swp
 .classpath
 .project
 .settings
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 .cache
-.cache-main
-*.swp
 .classpath
 .project
 .settings
--- a/indexing/README.md
+++ b/indexing/README.md
@@ -5,7 +5,7 @@
 
 Builds an elasticsearch index on an existing ElasticSearch instance, using one of the configurations defined in the indexing.conf file in resources/org/allenai/ari/indexing.
 
-To use this, you need to have a running instance of ElasticSearch (can be local or remote). As of this writing, the latest version is `1.7.2`. To install: Refer to http://joelabrahamsson.com/elasticsearch-101/ to get started, or use `brew install elasticsearch`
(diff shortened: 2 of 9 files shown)


matt-gardner:
The failed tests are because of a dependence on the datastore - is there a correct way to include tests that rely on the datastore?  I can't even pass in a FakeDatastore, or anything, because accessing the datastore uses static calls (Datastore.filePath() and Datastore.directoryPath()).

jkinkead:
This is an annoying feature of the default datastore client. If you delve into the datastore implementation, I believe there's a class you can mock which you could pass in as a constructor parameter.

jkinkead:
[This class](https://github.com/allenai/datastore/blob/master/datastore/src/main/scala/org/allenai/datastore/Datastore.scala#L40) should be mockable with Mockito, used occasionally in AI2.

matt-gardner:
I tried using Mockito, but it was giving me trouble, so I just went with the easier route of adding `getFromDatastore` methods that I could override in a test.

matt-gardner:
I haven't touched this in a long time, but several people used the changes I made here to create indexes.  Should I just merge this?

jkinkead:
Yeah, SGTM!

The circular dependency I was complaining about existing before this branch. :)

PR #167, opened 2016-01-05
String utilities from scholar

dirkgr:
I need these string utilities in science-parse, so I'm moving them from scholar to here.

When this is merged and released, I'll make a change to scholar using this instead of its own copy. There are two convenience functions though that are scholar specific that I didn't port over, year parsing and author name parsing. They will remain in scholar only.

This library already had a string implicit class. Because of conflicting names, I could not keep the old one around. I removed it, and put the functionality into the port from scholar. That makes this technically a source-code incompatible change that requires a major version upgrade. The change is in a tiny API though. I wrote that API, for my own use, and I'm not sure it's still used anywhere, so I felt that updating the major version number sends the wrong message. So instead, I bumped the minor number.


core/src/main/scala/org/allenai/common/StringImplicits.scala
```diff
@@ -1,115 +0,0 @@
-package org.allenai.common
```
dirkgr: This was moved into `StringUtils`.

core/src/main/scala/org/allenai/common/StringUtils.scala
```diff
@@ -0,0 +1,219 @@
+package org.allenai.common
```
dirkgr: None of this is new code, just copy and paste from `scholar`.

core/build.sbt
```diff
@@ -8,5 +8,6 @@ libraryDependencies ++= Seq(
   "redis.clients" % "jedis" % "2.7.2",
   Logging.logbackClassic,
   Logging.logbackCore,
-  Logging.slf4jApi
+  Logging.slf4jApi,
+  "org.apache.commons" % "commons-lang3" % "3.4"
```
markschaake: define this as a val in https://github.com/allenai/common/blob/4cf2a6194091cd4b596aa8f6db4e36f55d6e76f3/project/Dependencies.scala

core/src/main/scala/org/allenai/common/StringImplicits.scala
markschaake: so this is a breaking change. Do you want to define an alias in the package object and add the deprecated annotation? I think that might be a good solution.

in package.scala:
```scala
package org.allenai

package object common {
  @deprecated("Use 'import org.allenai.common.StringUtils._' instead", since = "1.0.21")
  val StringImplicits = StringUtils
}
```

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.0.21-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "1.1.1-SNAPSHOT"
```
markschaake: should this be "1.1.0-SNAPSHOT" ?

markschaake:
Done with my review. Looks good modulo comments. Feel free to merge / release when ready.

version.sbt
dirkgr: Oh yeah. There is another integer before 1. Forgot about that one :-)

core/build.sbt
dirkgr: done

dirkgr:
I saw on email that you had a comment about how this is a breaking change. I'm assuming you deleted those comments?

Like I said in the original writeup, I tried to avoid this problem, but the issue is that when you consume it you have to have two classes imported that are both called `StringImplicits`. Since they are supposed to be used implicitly, you don't get an opportunity to give a complete class path. We could work around this by naming the new one something else, but `StringImplicts` is a good name, and the benefit would be small, so I decided that a small breakage is the lesser evil.

markschaake:
> I saw on email that you had a comment about how this is a breaking change. I'm assuming you deleted those comments?

No, I didn't delete it. It is still there. The comment suggested adding an alias to the old `StringImplicits` in a package object to avoid breaking changes. See above

dirkgr:
Oh! I didn't see it despite looking for it. Sorry about that.

core/src/main/scala/org/allenai/common/StringImplicits.scala
dirkgr: That's not enough to be true source-code compatible. The old `StringImplicits` had members that could have been accessed, and the names of those members aren't the same as they are in `StringUtils`.

However, this way at least you'll get a useful warning telling you why it breaks. If you think that's good enough, I'll send a code review with that change.

PR #168, opened 2016-01-06
Useless release branch

dirkgr:


jkinkead:
The point of the release branch is to communicate to others that you're doing a release. It's also to give you a chance to review your commits so that you can ensure you're not pushing local-only changes to `master`.

I'm happy to do common releases for you if they're too burdensome. :P

dirkgr:
If you do it, I have to wait until you're around. When I do a common release, I want to consume it immediately somewhere else. I don't want to wait until you're around, and I certainly don't want to wait for a code review.

jkinkead:
Code reviews of this kind should be quick. They're also not the primary goal - although they give a good place for me to say things like: Since this branch is on your private fork, the release tag didn't get propagated to `allenai/common`, and you'll need to push it separately.

If you need to consume immediately, `publishLocal` still functions as a workaround.

If you want to update `sbt-plugins` with a new release process . . . and propagate to other ai2 projects . . . please be my guest. I'm not terribly attached to what we have, but there were reasonable reasons behind the choices made to get there.

dirkgr:
How many times have we had a problem where someone released something prematurely, and it broke other people?

`publishLocal` works great if all your code is on a single machine. It's a pain when you're running some of your stuff in EC2, and it's completely impossible if you want to consume the new thing in the S2 pipeline.

dirkgr:
Can I just bring `sbt release` back? I think that worked great.

jkinkead:
The problem of missing tags was a real one (although that continues to be one, per my previous comment). Likewise, the situation where we would publish an artifact and only push the `version.sbt` changes to the local branch happened frequently.

I don't think we had a situation where someone released a version that didn't match a tag on `master`, or released a version of `master` that had broken tests / code. However, given how few safeguards there were in place, it's impossible to say whether or not this ever happened.

In terms of the status quo, I think the big problem here is that the process is quite manual: You need to know a bunch of steps, and execute them in the correct order, for it to succeed. I think the tradeoff is reasonable - the process is quite well-documented. If you follow the "[Releasing new versions](https://github.com/allenai/common#releasing-new-versions)" instructions in the main README it takes very little time (and keeps you from making the mistake you made with your release).

I don't think we should go back to `sbt release`, since it's really easy to screw up - but we could try for a process that's a lot less manual.

PR #173, opened 2016-01-20
Upgrade sbt-plugins to 1.1.8

markschaake:
Also formatted files

@jkinkead this is the other half of https://github.com/allenai/sbt-plugins/pull/203


jkinkead:
LGTM.

PR #175, opened 2016-01-25
Refeactor Version.fromResource to use the classLoader.getResource

andrewlmurray:
Previously we were using Version.getClass.getResource vs Version.getClass.getClassLoader.getResource
which would prevent artifact.conf and git.conf from resolving properly when Version & artifact.conf
were not in the same jar.

### DIFF
--- a/core/src/main/scala/org/allenai/common/Version.scala
+++ b/core/src/main/scala/org/allenai/common/Version.scala
@@ -80,11 +80,18 @@ object Version {
     * @param  name  the sbt name, or artifact id, the calling code is in
     */
   def fromResources(org: String, name: String): Version = {
-    val pkg = "/" + org + "/" + name.replaceAll("-", "")
-    val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
-    val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-    require(artifactConfUrl != null, "Could not find artifact.conf in " + pkg + ".")
-    require(gitConfUrl != null, "Could not find git.conf in " + pkg + ".")
+    val prefix = s"$org/${name.replaceAll("-", "")}"
+
+    val artifactConfPath = s"$prefix/artifact.conf"
+    val gitConfPath = s"$prefix/git.conf"
+
+    val classLoader = this.getClass.getClassLoader
+
+    val artifactConfUrl = classLoader.getResource(artifactConfPath)
+    val gitConfUrl = classLoader.getResource(gitConfPath)
+
+    require(artifactConfUrl != null, s"Could not find $artifactConfPath")
+    require(gitConfUrl != null, s"Could not find $gitConfPath")
 
     val artifactConf = ConfigFactory.parseURL(artifactConfUrl)
     val gitConf = ConfigFactory.parseURL(gitConfUrl)


markschaake:
LGTM

core/src/main/scala/org/allenai/common/Version.scala
```diff
@@ -80,11 +80,18 @@ object Version {
     * @param  name  the sbt name, or artifact id, the calling code is in
     */
   def fromResources(org: String, name: String): Version = {
-    val pkg = "/" + org + "/" + name.replaceAll("-", "")
-    val artifactConfUrl = this.getClass.getResource(pkg + "/artifact.conf")
-    val gitConfUrl = this.getClass.getResource(pkg + "/git.conf")
-    require(artifactConfUrl != null, "Could not find artifact.conf in " + pkg + ".")
-    require(gitConfUrl != null, "Could not find git.conf in " + pkg + ".")
+    val prefix = s"$org/${name.replaceAll("-", "")}"
```
jkinkead: How does removing the leading slash change the resource finding behavior?

jkinkead: Clearly: In what way does it change the behavior?

jkinkead:
I'm concerned that you would be loading versions from two different jars. Wouldn't that totally invalidate the notion of a version for a single artifact?

andrewlmurray:
@jkinkead this particular case is the scholar pipeline. The assembly dependencies (which contains Version.scala) and the application code (which ends up containing artifact) are in two separate jars, but always come from the same git sha.

jkinkead:
I still don't understand why the application jar is missing its `.conf` file. It seems like this is a bug in our plugin, not a bug in how we load the `Version` data.

andrewlmurray:
@jkinkead FYI the application jar is not missing its application.conf, rather the Version class is in a separate jar. 

dependencies jar
----------------
some/package/Version.class
.... tons of other stuff ...

assembly.jar
--------------------
org.allenai.s2/pipeline/artifact.conf

jkinkead:
But this is the case for us as well - we have the `.conf` files in our application jars, and the version class in another jar, and have no trouble resolving these.

Are you using some sort of application container with multiple class loaders or something?

andrewlmurray:
@jkinkead yeah right you are (sorry this is almost impossible to reproduce without actually deploying) - the behavior doesn't seem to be any different after this patch and it seems to be due to spark classLoader funk. 

I had tested my change by running an inlined new version of Version.fromResources. Of course this used my own classes classLoader, which made it look like my patch fixed the issue - in reality the fix was to use the same classLoader as a class going into our assembly.jar. 

A class in the main pipeline (which ends up in assembly.jar) does the following

```scala
var summary = Summary(SummaryRunMetadata(
    inputSnapshotUri,
    LocalDate.now,
    Version.fromResources("org.allenai.s2", "pipeline").git.sha1
))
```

This, run locally works / in the console all works

```scala
Version.fromResources("org.allenai.s2", "pipeline")
res0: org.allenai.common.Version = Version(GitVersion(47c46b83382daefb3c649817932214100055d2c0,1453763209000,Some(http://github.com/allenai/scholar)),2014.09.02-5-SNAPSHOT,None)
```

Deployed in spark I still get this assertion:

```
java.lang.IllegalArgumentException: requirement failed: Could not find org.allenai.s2/pipeline/artifact.conf
```

The dependency jar contains artifact / git confs for the dependencies
```
./org.allenai/scienceparse/artifact.conf
./org.allenai.common/commoncore/artifact.conf
./org.allenai.nlpstack/nlpstackcore/artifact.conf
./org.allenai.nlpstack/nlpstacksegment/artifact.conf

```
The application.jar contains the artifact/git confs for the project I need the data from.

andrewlmurray:
(Either way this patch should be safe  - but it won't fix my problem)

jkinkead:
I'm glad that my understanding of how this worked wasn't totally crazy - but I wish your problem was fixed!

It seems like we should add a method to `Version` that takes a `ClassLoader` instance. That way, you can pass in the loader for the main application, which should be able to read from the jar it came in.

andrewlmurray:
Yes - I side-stepped having to do the release song / dance by just doing the lookup manually, but yes I think parameterizing the Version by a Classloader (defaulting to Version's own) makes total sense.

PR #176, opened 2016-01-25
Use classloader for version

andrewlmurray:


andrewlmurray:
Release song and dance!

markschaake:
Assigning to @jkinkead since he has questions in https://github.com/allenai/common/pull/175

andrewlmurray:
Sounds good! @jkinkead I responded to you in #175

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
jkinkead: This should go in your `~/.gitignore`.

jkinkead:
I'm slightly concerned that this will introduce bugs into downstream dependencies, since it changes how `.conf` files are found - but honestly, that's probably already classpath-order dependent, and we don't make an effort to sort that in a sensible way.

I'm also worried that this is a workaround for another bug, but it ought to be harmless either way.

PR #180, opened 2016-03-11
Add ability to pass `timeout` parameter to JsonQueryCache.

danxmoran:
While poking around the handler we use in Aristo to cache solver responses, I noticed the [timeout argument](https://github.com/allenai/aristo/blob/master/controller/src/main/scala/org/allenai/ari/controller/solverclient/SolverCacheHandler.scala#L49) is being completely ignored. The correct behavior would be to use the configured timeout to construct the JSON cache in the handler's companion object, which in turn would use the timeout when constructing its `JedisPool`.


jkinkead:
LGTM!

PR #182, opened 2016-04-06
Add guice module for JsonQueryCache, and improve test usability

danxmoran:
@jkinkead I think I set up the guice module properly (I set it up assuming we didn't want the guice libraries to creep into the `core` subproject via annotations on the cache class).  Let me know if there's anything I should change!


danxmoran:
I should say, this PR is a step towards moving the wumpus client (and eventual ermine and lucene clients) onto a common caching mechanism, since all of those projects use guice.

jkinkead:
Putting this in the `guice` subproject is really strange.

What I'd actually prefer is if the caching library were in its own subproject (instead of `core`), if we're trying to keep dependencies clean. I think we're a lot more likely to generally use Guice than Redis. :)

danxmoran:
Sub-project created!

danxmoran:
Cleaned up the `build.sbt`s while I was at it, too.

build.sbt
```diff
@@ -22,35 +22,34 @@ lazy val buildSettings = Seq(
   bintrayPackage := s"${organization.value}:${name.value}_${scalaBinaryVersion.value}"
 )
 
-lazy val testkit = Project(
-  id = "testkit",
-  base = file("testkit"),
-  settings = buildSettings
-).enablePlugins(LibraryPlugin)
+lazy val caching = Project(id = "caching", base = file("caching"))
```
jkinkead: Nitpick: The subproject name should match the package name - either both should be `cache`, or both should be `caching`.

caching/src/test/scala/org/allenai/common/cache/JsonQueryCacheSpec.scala
```diff
@@ -1,27 +1,24 @@
 package org.allenai.common.cache
 
-import java.io.File
-import org.allenai.common.{ GitVersion, Version, cache }
 import org.allenai.common.testkit.UnitSpec
-import org.scalatest.BeforeAndAfterAll
-import spray.json.DefaultJsonProtocol
-import sys.process._
+
+import com.fiftyonred.mock_jedis.MockJedisPool
```
jkinkead: Cool! I don't the original author found this library. We had something in place originally that was buggy (either the included mock library, or something else . . .)

jkinkead:
Just one comment re: naming. Otherwise LGTM!

danxmoran:
Updated the sub-project name.  Will merge on green!

PR #183, opened 2016-04-08
Release 1.2.0

danxmoran:
Decided to bump the minor version since moving the cache out of common-core could cause headaches for anyone caught off-guard.


jkinkead:
Yeah, that's pretty much the definition of a non-backwards-compatible change :)

PR #184, opened 2016-04-11
Fix JsonQueryCacheModule, and add tests.

danxmoran:
In addition to the tests here, I've verified it works in the wumpus project.


cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
```diff
@@ -4,14 +4,25 @@ import org.allenai.common.guice.ConfigModule
...
-import com.typesafe.config.{ Config, ConfigFactory }
+import com.typesafe.config.Config
 import redis.clients.jedis.{ JedisPool, JedisPoolConfig }
+import spray.json.JsonFormat
+
+/** A guice module for binding a `JsonQueryCache` instance with a parameterized value type. */
+class JsonQueryCacheModule[V](config: Config)(
+  implicit
+  jsonFormat: JsonFormat[V],
+  manifest: Manifest[V]
```
danxmoran: The `Manifest` is what was needed to get the parameterized binding to work properly.  A little confusing since most of the existing Scala docs say `Manifest` is deprecated, but if you look in `Predef` you see it's been "un-deprecated".

jkinkead: I think there was something like `ClassTag` that was preferred at one point or another.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
```diff
@@ -4,14 +4,25 @@ import org.allenai.common.guice.ConfigModule
...
-class JsonQueryCacheModule(config: Config = ConfigFactory.empty) extends ConfigModule(config) {
   override def configName: String = "cache.conf"
 
   override def bindingPrefix: Option[String] = Some("redis")
 
+  override def configureWithConfig(fullConfig: Config): Unit = {
+    bind[JsonFormat[V]].toInstance(jsonFormat)
+  }
+
   @Provides def provideJedisPool(
```
jkinkead: If you mark this `@Singleton`, it'll fix the new-pool problem.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
```diff
@@ -4,14 +4,25 @@ import org.allenai.common.guice.ConfigModule
...
+  jsonFormat: JsonFormat[V],
+  manifest: Manifest[V]
+) extends ConfigModule(config) {
 
-class JsonQueryCacheModule(config: Config = ConfigFactory.empty) extends ConfigModule(config) {
   override def configName: String = "cache.conf"
 
   override def bindingPrefix: Option[String] = Some("redis")
 
+  override def configureWithConfig(fullConfig: Config): Unit = {
```
jkinkead: Another simpler option is to create a new factory class for the query cache, and inject that. The factory class would have a `getInstance[V]` method that takes the `JsonFormat[V]`, and you won't need to deal with manifests at all.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
danxmoran: It'll stop errors from being thrown, but won't it *really* enforce a limitation of a single pool per injector then?  Or does `@Singleton` mean something different in guice?

jkinkead:
OK, I understand this a lot better now. Two suggestions that should make the module simpler (no type parameters / manifests) and fix the pool problem.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
jkinkead: The Guice `@Singleton` annotation means exactly that: Only one instance will be created per injector.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
danxmoran: Have you ever tried working with guice's auto-factory-generation mechanisms: https://stackoverflow.com/questions/8976250/how-to-use-guices-assistedinject?  Are they worth my time to explore for this?

danxmoran: Actually, if we're going to inject a factory to code that needs a cache instead of a cache itself, why bother using a guice module?  Could we just scrap this and add a `JsonQueryCache` object with a `fromConfig` factory method?

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
danxmoran: Wouldn't that then force all caches bound to an injector to point at the same Redis host?  I thought that wasn't what we wanted.

danxmoran: To me, the ideal might be to move the "clientPrefix" parameter we currently expect to find in config up to be an argument for the module's constructor.  Then we can use it as a binding prefix for everything else.

danxmoran:
@jkinkead I decided to ditch the cache module entirely in favor of a simple `fromConfig` factory method on the cache's companion object.  I think this is a lot simpler / cleaner overall.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCacheModule.scala
jkinkead: Yep! If you inject the connection pool instead. :)

cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
```diff
@@ -1,10 +1,29 @@
...
+import com.typesafe.config.Config
 import redis.clients.jedis.{ Jedis, JedisPool, JedisPoolConfig, Protocol }
 import spray.json._
 
+object JsonQueryCache {
+  /** Factory method for creating a cache instance from config.
+    * The config must have keys for `hostname` and `clientPrefix`. It may also optionally have
+    * keys for `port` and `timeoutMillis`; if not given, these values are set to the Jedis defaults.
+    */
+  def fromConfig[V](config: Config)(implicit jsonFormat: JsonFormat[V]): JsonQueryCache[V] = {
```
jkinkead: I think it would be nice to have a singleton pool (possibly injected with a module). This still has multiple pools.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
```diff
@@ -25,7 +41,10 @@ class JsonQueryCache[V: JsonFormat] @Inject() (
     * become part of the cache key (prepended to the actual query)
     */
   def this(redisHostName: String, redisHostPort: Int, redisTimeout: Int, clientPrefix: String) =
-    this(new JedisPool(new JedisPoolConfig, redisHostName, redisHostPort, redisTimeout), clientPrefix)
+    this(
+      new JedisPool(new JedisPoolConfig, redisHostName, redisHostPort, redisTimeout),
```
jkinkead: I really don't like multiple constructors on objects. It makes initialization a lot harder to get right. :/

cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
```diff
@@ -1,10 +1,29 @@
...
+  def fromConfig[V](config: Config)(implicit jsonFormat: JsonFormat[V]): JsonQueryCache[V] = {
+    // Required fields.
+    val hostname: String = config[String]("hostname")
+    val clientPrefix: String = config[String]("clientPrefix")
+
+    // Optional overrides for Jedis defaults.
+    val port: Int = config.get[Int]("port") getOrElse Protocol.DEFAULT_PORT
+    val timeoutMillis: Int = config.get[Int]("timeoutMillis") getOrElse Protocol.DEFAULT_TIMEOUT
+
+    new JsonQueryCache[V](hostname, port, timeoutMillis, clientPrefix)
```
jkinkead: You should call the primary constructor, with a connection pool you create here.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
jkinkead: Rethinking: Never mind. This is fine.

jkinkead:
One comment about the multiple constructors that exist before your changes; one comment about preferring the primary constructor.

cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
danxmoran: I've been resisting the urge to remove these in case somebody's using them in another project, but I always forget that their code won't break unless they explicitly bump the version number.

danxmoran:
Comments addressed :)

jkinkead:
LGTM! You're still bumping the `scala-guice` version, but I assume that's fine.

Merge when you're happy!

PR #185, opened 2016-04-11
Make default config in ConfigModule optional.

danxmoran:
@jkinkead one more thing I want to sneak in before cutting a new release of common.  I find the `ConfigModule` insanely convenient, but dislike that it requires specifying a `configName` for default values.  Our current approach of trying to load a dummy file `module.conf` that may or may not exist only seems to work because of the peculiarities of the `ConfigFactory` implementation, and strikes me as major code smell.

This PR makes the field `configName` optional.


jkinkead:
Awesome, this looks great.

I'm also glad that this config module is proving useful to someone other than myself. :)

PR #187, opened 2016-04-14
Add client helpers for spray to webapp.

danxmoran:
Consolidation of patterns used in aristo and ermine.

### DIFF
--- a/webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
+++ b/webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
@@ -127,7 +127,7 @@ object SprayClientHelpers {
     val clientConnectionSettings = ClientConnectionSettings(system).copy(
       requestTimeout = requestTimeout,
       // Amount of time an idle HTTP connection will be held open before being closed.
-      idleTimeout = Duration.Inf,
+      idleTimeout = requestTimeout * 2,
       connectingTimeout = connectionTimeout,
       userAgentHeader = connectorId map { `User-Agent`(_) }
     )
@@ -139,7 +139,7 @@ object SprayClientHelpers {
         HostConnectorSettings(system).copy(
           // Amount of time one of spray's HostConnector actors will sit idle before terminating
           // itself.
-          idleTimeout = Duration.Inf,
+          idleTimeout = requestTimeout * 2,
           maxConnections = maxConnections,
           connectionSettings = clientConnectionSettings
         )
--- a/webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
+++ b/webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
@@ -1,0 +1,83 @@
+package org.allenai.common.webapp
+
+import akka.actor.ActorSystem
+import akka.io.IO
+import akka.pattern.ask
+import akka.util.Timeout
+import spray.can.Http
+import spray.can.Http.HostConnectorSetup
+import spray.can.client.{ ClientConnectionSettings, HostConnectorSettings }
+import spray.http.HttpHeaders.`User-Agent`
+import spray.http.{ HttpRequest, HttpResponse }
+
+import java.net.URL
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+object SprayClientHelpers {
+  /** Send an HTTP request through spray using a dedicated `HostConnectorSetup`, and process the
+    * response using the given function. This gives you much more control over how and when your
+    * request is sent over the wire / when it times out than does use of `sendReceive`.
+    * @param request the `HttpRequest` object to send over the wire
+    * @param connectorSetup the `HostConnectorSetup` object defining connection and timeout
+    * information for your request
+    * @param parseResponse a function from `HttpResponse` to a generic value `T` you want to
+    * extract from your request's response
+    */
+  def sendRequest[T](
+    request: HttpRequest,
+    connectorSetup: HostConnectorSetup
+  )(parseResponse: HttpResponse => T)(implicit actorSystem: ActorSystem): Future[T] = {
+    import actorSystem.dispatcher
+
+    // Spray requires a top-level timeout for sending requests into its infrastructure even
+    // though it has internal timeouts built-in. We set this ridiculously high timeout here so we
+    // never have to deal with uninformative `AskTimeoutExceptions`.
+    implicit val askTimeout = Timeout(Int.MaxValue.millis)
+
+    IO(Http).ask((request, connectorSetup)) map {
+      case response: HttpResponse => parseResponse(response)
+    }
+  }
+
+  /** Override spray's default settings for sending requests with the given timeouts.
+    * @param url the URL of the remote host to communicate with
+    * @param connectionTimeout the timeout to use when establishing a remote connection to the
+    * remote host
+    * @param requestTimeout the timeout to use when waiting for a response from the remote host
+    * @param maxConnections the maximum number of connections to the remote host that will be held
+    * open at a time by the returned connector
+    * @param connectorId an optional unique ID to use as the User-Agent header for requests to
+    * this host. If you want to maintain multiple host connector pools to a single remote host
+    * (for example, to prevent long-running requests from interfering with quick requests),
+    * setting this field to different values for each will prevent spray from sharing a connector
+    * between the different types of requests.
+    */
+  def getConnectionSetup(
+    url: URL,
+    connectionTimeout: FiniteDuration,
+    requestTimeout: FiniteDuration,
+    maxConnections: Int,
+    connectorId: Option[String] = None
+  )(implicit system: ActorSystem): HostConnectorSetup = {
+    val clientConnectionSettings = ClientConnectionSettings(system).copy(
+      requestTimeout = requestTimeout,
+      idleTimeout = requestTimeout * 2,
+      connectingTimeout = connectionTimeout,
+      userAgentHeader = connectorId map { `User-Agent`(_) }
+    )
+
+    HostConnectorSetup(
+      host = url.getHost,
+      port = url.getPort,
+      settings = Some(
+        HostConnectorSettings(system).copy(
+          idleTimeout = requestTimeout * 2,
+          maxConnections = maxConnections,
+          connectionSettings = clientConnectionSettings
+        )
+      )
+    )
+  }
+}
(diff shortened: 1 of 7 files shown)


jkinkead:
We don't currently run it tests for common. I'm going to enable those, and see what this branch does.

I *think* semaphore lets us open ports, but sometimes that's not allowed in a CI system.

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,83 @@
...
+  def sendRequest[T](
+    request: HttpRequest,
+    connectorSetup: HostConnectorSetup
+  )(parseResponse: HttpResponse => T)(implicit actorSystem: ActorSystem): Future[T] = {
+    import actorSystem.dispatcher
+
+    // Spray requires a top-level timeout for sending requests into its infrastructure even
+    // though it has internal timeouts built-in. We set this ridiculously high timeout here so we
+    // never have to deal with uninformative `AskTimeoutExceptions`.
+    implicit val askTimeout = Timeout(Int.MaxValue.millis)
```
jkinkead: This just reminds me how little I like `spray-http`. :)

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,83 @@
...
+  def getConnectionSetup(
+    url: URL,
+    connectionTimeout: FiniteDuration,
+    requestTimeout: FiniteDuration,
+    maxConnections: Int,
+    connectorId: Option[String] = None
+  )(implicit system: ActorSystem): HostConnectorSetup = {
+    val clientConnectionSettings = ClientConnectionSettings(system).copy(
+      requestTimeout = requestTimeout,
+      idleTimeout = requestTimeout * 2,
```
jkinkead: I'd make this a parameter to the method, but with a default value of `requestTimeout * 2`. Often you want an infinite idle timeout.

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,83 @@
...
+    * @param maxConnections the maximum number of connections to the remote host that will be held
+    * open at a time by the returned connector
+    * @param connectorId an optional unique ID to use as the User-Agent header for requests to
+    * this host. If you want to maintain multiple host connector pools to a single remote host
+    * (for example, to prevent long-running requests from interfering with quick requests),
+    * setting this field to different values for each will prevent spray from sharing a connector
+    * between the different types of requests.
+    */
+  def getConnectionSetup(
+    url: URL,
```
jkinkead: I'd take the host & port instead, for clarity.

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,83 @@
...
+import spray.can.client.{ ClientConnectionSettings, HostConnectorSettings }
+import spray.http.HttpHeaders.`User-Agent`
+import spray.http.{ HttpRequest, HttpResponse }
+
+import java.net.URL
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+object SprayClientHelpers {
```
jkinkead: It would be cool to have examples of how to use in an object-level scaladoc.

jkinkead:
A few comments!

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
danxmoran: I'd need to use an awkward curried parameter list to make the default idle timeout depend on the value of `requestTimeout`: http://stackoverflow.com/a/3090906

I've wondered for awhile about this: what's the value of setting an idle timeout if you've already got a tighter request timeout?

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,83 @@
...
+      connectingTimeout = connectionTimeout,
+      userAgentHeader = connectorId map { `User-Agent`(_) }
+    )
+
+    HostConnectorSetup(
+      host = url.getHost,
+      port = url.getPort,
+      settings = Some(
+        HostConnectorSettings(system).copy(
+          idleTimeout = requestTimeout * 2,
```
danxmoran: Since we're discussing idle timeouts, is it a potential problem that we conflate *this* idle timeout with the one above?  The one above is "The time after which an idle connection will be automatically closed." (quoting spray docs), while this one is "The time after which an idle `HttpHostConnector` (without open connections) will automatically terminate itself."

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
jkinkead: Aside: I don't like `spray-client` anymore.

Usually when you're using an HTTP client to talk repeatedly to a single host you'll want to use a connection pool with persistent HTTP connections. That way you can avoid the connection overhead with each request. Usually when I hear "idle" associated with a connection, I assume it affects the amount of time a connection can remain in the pool before being dropped.

It could also mean the amount of time a request will wait for more bytes on a socket before giving up and unlistening. [The scaladoc on ClientConnectionSettings](https://github.com/spray/spray/blob/76ab89c25ce6d4ff2c4b286efcc92ee02ced6eff/spray-can/src/main/scala/spray/can/client/ClientConnectionSettings.scala#L25) doesn't say which it is.

I'd say stick with an unlimited timeout in case it's the former, given that we have a reasonable request timeout which makes it meaningless if it's the latter.

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
jkinkead: I think both should be unlimited, provided we have an upper-bound on the number of connections created.

danxmoran: A good example of where this distinction has bit us in the past: in the `SolverReplicas` we used to build a `HostConnectorSetup` at startup and then send it to `IO(Http)` to get a `HostConnector`, to which we could then send HTTP requests directly.  We had the `HostConnectorSetup.idleTimeout` set to `2 * requestTimeout` because we were treating it the same as the idle timeout for TCP connections.  After sitting idle all night following a nightly deploy, the controller became unusable because all of the `HostConnectors` created at startup terminated themselves.

danxmoran: Responding to your comment: ok, will do!

danxmoran:
Comments addressed!

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
danxmoran: I've read that [Play's WS library](https://www.playframework.com/documentation/2.5.x/ScalaWS) is pretty good.  It's possible to use without depending on the whole of playframework.

danxmoran: The downside of course being that it apparently uses an entirely different JSON-parsing library -_-

jkinkead: That looks a lot better than `spray`. I'm not sure I trust anything coming out of Play, though. :P

[OK HTTP](http://square.github.io/okhttp/) is what I'd first check out.

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
```diff
@@ -0,0 +1,149 @@
...
+    // though it has internal timeouts built-in. We set this ridiculously high timeout here so we
+    // never have to deal with uninformative `AskTimeoutExceptions`.
+    implicit val askTimeout = Timeout(Int.MaxValue.millis)
+
+    IO(Http).ask((request, connectorSetup)) map {
+      case response: HttpResponse => parseResponse(response)
+    }
+  }
+
+  /** Override spray's default settings for sending requests with the given timeouts.
```
jkinkead: You should note the infinite idle timeouts. I'm not sure how much this matters exactly.

These are magically cached somewhere inside the actor system, right? We're not going to create infinite connection pools or whatever?

jkinkead:
LGTM apart from one piece of doc that might be nice to have!

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
jkinkead: It's really easy to parse a string or similar object with `spray-json`.

We could also port to my `reming` library if we're feel ambitious. Or just use protobuffs. :P

danxmoran: I'll port the whole thing to use protobuf and grpc if you commit to maintaining a proto-to-Scala compiler :P

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
danxmoran: Setting the idle timeout of the `HostConnectorSettings` to infinite means that every `Actor` spray creates will live until the actor system dies, even if it's sitting around doing nothing for hours.  The total number of actors created will be bounded by the `maxConnections` argument, though, and I believe they're created lazily, so I'm not sure how much of a practical issue this could be.

danxmoran:
I reset the idle timeouts back to `2 * requestTimeout` because infinite timeouts were making me nervous.  Regardless, I added a mention in the docs!

danxmoran:
One gotcha: looks like we're still not running ITs in Semaphore.  Is that on purpose?  Does Semaphore not allow opening ports after all?

webapp/src/main/scala/org/allenai/common/webapp/SprayClientHelpers.scala
jkinkead: Weekend TODOs:
- [ ] mow lawn
- [ ] write pb -> scala library
- [ ] weed garden

jkinkead:
We are now! :)

jkinkead:
LGTM - merge on green!

PR #189, opened 2016-04-27
Consolidate a few commonly used utility methods

ashish33:
@jkinkead Following up on our chat, moving some handy utility methods (useful at least for multiple Aristo solvers) to this common location.


ashish33:
The build seems to be failing in a part of the code this PR doesn't really touch:
```
[warn] ...StringUtilsSpec.scala:4: imported `StringImplicits' is permanently hidden by definition of object StringImplicits in package common
[error] ...StringUtilsSpec.scala:8: value titleCaseIfAllCaps is not a member of String
```
Ideas?

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import java.io._
+
+import scala.collection.JavaConverters._
+import scala.io.{ BufferedSource, Codec, Source }
+
+/** Various convenient utilities for reading files and resources. */
+object FileUtils extends Logging {
+
+  /** Get a file as a buffered source. */
```
jkinkead: This doesn't do anything other than delegate to the `Source` implementation.

Instead of using this method, you should do something like:
```scala
// Use the optional codec if it's provided; else, use the default codec.
val localCodec = myOptionCodec.getOrElse(implicitly[Codec])
val source = Source.fromFile(file)(localCodec)
```
or with implicits:
```scala
// Implicit codec; really should only be used if it's shared in a class.
implicit val localCodec = myOptionCodec.getOrElse(implicitly[Codec])
val source = Source.fromFile(file)
```

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+  def getFileAsSource(file: File, codecOpt: Option[Codec] = None): BufferedSource = {
+    codecOpt match {
+      case Some(codec) => Source.fromFile(file)(codec)
+      case None => Source.fromFile(file)
+    }
+  }
+
+  /** Get a file as a sequence of lines. */
```
jkinkead: You should note that this is a non-lazy sequence.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+  def getFileAsLines(file: File, codecOpt: Option[Codec] = None): Seq[String] = {
+    logger.debug(s"Loading file ${file.getName}")
+    Resource.using(getFileAsSource(file, codecOpt)) { input =>
+      val lines = input.getLines().toVector // convert to vector to force stream to be processed
+      logger.trace(s"lines:\n\t${lines.mkString("\n\t")}")
```
jkinkead: This should really go away.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+      lines
+    }
+  }
+
+  /** Read a CSV file into a sequence (rows) of sequence (columns) of strings. */
+  def getCSVContentFromFile(file: File): Seq[Seq[String]] = {
+    logger.debug(s"Loading CSV file ${file.getName}")
+    val csvReader = new CSVReader(new FileReader(file))
+    Resource.using(csvReader)(_.readAll.asScala.map(_.toSeq))
```
jkinkead: You should use `toVector` here.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
+package org.allenai.common
+
+import _root_.au.com.bytecode.opencsv.CSVReader
```
jkinkead: Why the `_root_` here? This shouldn't be needed . . .

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /** Read a CSV file into a sequence (rows) of sequence (columns) of strings. */
+  def getCSVContentFromFile(file: File): Seq[Seq[String]] = {
+    logger.debug(s"Loading CSV file ${file.getName}")
+    val csvReader = new CSVReader(new FileReader(file))
+    Resource.using(csvReader)(_.readAll.asScala.map(_.toSeq))
+  }
+
+  /** Get a resource file as a Stream. Caller is responsible for closing this stream. */
+  def getResourceAsStream(name: String): BufferedInputStream = {
```
jkinkead: I don't think you want these helpers. It obfuscates which type of resource you're loading.

More importantly, you're using the current class to locate the resource, which isn't going to work unless the resource is in the `org.allenai.common` package.

If you want have this take a `Class` instance as a parameter, and clearly document what it's returning, that would be OK.

All in all, I recommend against using resources, because of issues like this. :)

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,73 @@
...
+    new BufferedInputStream(getClass.getClassLoader.getResourceAsStream(name))
+  }
+
+  /** Get a resource file as a Reader. Caller is responsible for closing this reader. */
```
jkinkead: You aren't providing a `Codec` here. You should (and pass it through to the `InputStreamReader` constructor as a `Charset`).

jkinkead: (probably as an implicit parameter).

core/src/main/scala/org/allenai/common/MathUtils.scala
```diff
@@ -0,0 +1,10 @@
+package org.allenai.common
+
+/** Various convenient utilities for math operations. */
+object MathUtils {
+
+  /** Round a Double to k decimal digits. */
```
jkinkead: Maybe take an optional rounding mode?

project/Dependencies.scala
```diff
@@ -23,6 +23,8 @@ object Dependencies extends CoreDependencies {
   def nlpstack(component: String) = ("org.allenai.nlpstack" % s"nlpstack-${component}_2.11" % "1.6")
     .exclude("commons-logging", "commons-logging")
 
+  val openCsv = "net.sf.opencsv" % "opencsv" % "2.1"
```
jkinkead: I'm not sure we want this in the main common package . . . but it's probably fine.

core/src/main/scala/org/allenai/common/DatastoreUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+
+/** Various convenient utiltiies for accessing the Datastore. */
+object DatastoreUtils extends Logging {
+  /** Get a datastore file as a buffered Source. Caller is responsible for closing this stream. */
+  def getDatastoreFileAsSource(
+    datastoreName: String,
+    group: String,
+    name: String,
+    version: Int
+  ): BufferedSource = {
```
jkinkead: Should take implicit codec.

core/src/main/scala/org/allenai/common/DatastoreUtils.scala
```diff
@@ -0,0 +1,53 @@
...
+    logger.debug(s"Loading file from $datastoreName datastore: $group/$name-v$version")
+    val file = Datastore(datastoreName).filePath(group, name, version).toFile
+    Source.fromFile(file)
+  }
+
+  /** Get a datastore file as a buffered Source. Caller is responsible for closing this stream. */
+  def getDatastoreFileAsSource(config: TypesafeConfig): BufferedSource = {
```
jkinkead: + implicit code.

jkinkead:
I'm looking at the failure now. It might be an existing test issue; I'll send a PR if I can fix.

jkinkead:
I'm rebuilding now with updated settings. Let's see if it completes!

jkinkead:
Hrm, still the same weird error.

I can build fine locally - can you build without error also?

jkinkead:
I also just kicked off a build from `master` on this project to see if semaphore can repro it not on your pull.

ashish33:
It builds fine but `core/test` fails for me locally.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Dropping this method and instead just using `Source.fromFile(file)` with implicit `codec` where this method was used.

jkinkead:
Ugh. That's weird - I couldn't reproduce locally myself. Let me try with a clean build.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Good point.  Done here and elsewhere.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Indeed!  Dropped.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: :-)  Dropped.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Hmm.  Got several places where `tableilp` in Aristo uses resources.  Changed to have this take a `Class` parameter.  Better?

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Added `Codec` as an implicit parameter to all applicable methods.  I'm not very familiar with implicits though, so please check if the changes make sense?

core/src/main/scala/org/allenai/common/MathUtils.scala
ashish33: Done.  Taking `BigDecimal.RoundingMode.Value` as an optional parameter.  OK or would it be better to define our own `BigDecimal`-independent rounding type?

ashish33:
Incorporated many nice suggestions!  But not fully sure about functionality, especially around implicits. So please check.

Semaphore is still unhappy :-(

ashish33:
Dropping `DatastoreUtils` from here to avoid circular dependency between `common` and `datastore` projects.  Will likely include it in the `datastore` project later.

Hopefully this makes Semaphore happy!

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+import scala.collection.JavaConverters._
+import scala.io.{ BufferedSource, Codec, Source }
+
+import java.nio.charset.Charset
+
+/** Various convenient utilities for reading files and resources. */
+object FileUtils extends Logging {
+
+  /** Get a file as a non-lazy sequence of lines. */
+  def getFileAsLines(file: File)(codec: Codec): Seq[String] = {
```
jkinkead: Second parameter list should be marked `implicit`.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+    logger.debug(s"Loading file ${file.getName}")
+    // use toVector to force stream to be processed
+    Resource.using(Source.fromFile(file)(codec))(_.getLines().toVector)
+  }
+
+  /** Read a CSV file as a non-lazy sequence (rows) of sequence (columns) of strings. */
+  def getCSVContentFromFile(file: File): Seq[Seq[String]] = {
+    logger.debug(s"Loading CSV file ${file.getName}")
+    val csvReader = new CSVReader(new FileReader(file))
```
jkinkead: This method should take a codec & pass in the charset to the `FileReader`.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+    Resource.using(csvReader)(_.readAll.asScala.map(_.toVector))
+  }
+
+  /** Get a resource file for a given class as a Stream. Caller is responsible for closing this
+    * stream.
+    */
+  def getResourceAsStream(name: String, c: Class[_]): BufferedInputStream = {
```
jkinkead: `clazz` is the conventional name for `Class` instances. ;)

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+    new BufferedInputStream(c.getClassLoader.getResourceAsStream(name))
+  }
+
+  /** Get a resource file for a given class as a Reader. Caller is responsible for closing this
+    * reader.
+    */
+  def getResourceAsReader(name: String, c: Class[_])(codec: Codec): BufferedReader = {
+    val localCharset = codec.asInstanceOf[Charset]
```
jkinkead: `Codec` does not extend `Charset` - this will throw an exception!

You should use the `charSet` value `Codec` contains.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,66 @@
...
+    * stream.
+    */
+  def getResourceAsStream(name: String, c: Class[_]): BufferedInputStream = {
+    new BufferedInputStream(c.getClassLoader.getResourceAsStream(name))
+  }
+
+  /** Get a resource file for a given class as a Reader. Caller is responsible for closing this
+    * reader.
+    */
+  def getResourceAsReader(name: String, c: Class[_])(codec: Codec): BufferedReader = {
```
jkinkead: Mark second parameter list as `implicit`, here and in remaining functions.

project/Dependencies.scala
```diff
@@ -12,7 +12,7 @@ object Dependencies extends CoreDependencies {
 
   val commonsIO = "commons-io" % "commons-io" % "2.4"
 
-  val datastore = "org.allenai" %% "datastore" % "2015.04.02-0"
+  val datastore = "org.allenai" %% "datastore" % "1.0.0"
```
jkinkead: Remove.

jkinkead:
A few more comments.

Maybe a unit test to test these would be good?

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Done, here and elsewhere.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: Changed to `new CSVReader(new InputStreamReader(new FileInputStream(file), codec.charSet))` as `FileReader` doesn't like to take a `CharSet`.

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: `clazz` it is!

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: done

project/Dependencies.scala
ashish33: Keeping, as `indexing/.../BuildCorpusIndex.scala` currently depends on it.

ashish33:
Added unit tests.

core/src/main/scala/org/allenai/common/FileUtils.scala
```diff
@@ -0,0 +1,68 @@
+package org.allenai.common
+
+import au.com.bytecode.opencsv.CSVReader
+
+import java.io._
```
jkinkead: You should import specific classes.

project/Dependencies.scala
jkinkead: Can you add a TODO regarding the circular dependency from the indexing utils library? This needs to be broken.

jkinkead:
A couple of small comments - otherwise LGTM!

core/src/main/scala/org/allenai/common/FileUtils.scala
ashish33: At one point yesterday there were too many to explicitly specify, but now seems OK with only 5 classes.  Changing.

project/Dependencies.scala
ashish33: Done, as `TODO(jkinkead)` as this is beyond my expertise :-)

ashish33:
Addressed comments.  Good to merge?

jkinkead:
Yep, LGTM!

PR #190, opened 2016-04-27
Add a binding for ActorRefFactory.

jkinkead:
Suh guicey.

Sometimes all you need is the `ActorRefFactory`, and not the full `ActorSystem`. spray-routing's `compressResponseIfRequested` directive is a good example of an API which requires the factory, but not the full system.

### DIFF
--- a/guice/src/main/scala/org/allenai/common/guice/ActorSystemModule.scala
+++ b/guice/src/main/scala/org/allenai/common/guice/ActorSystemModule.scala
@@ -1,11 +1,12 @@
 package org.allenai.common.guice
 
-import akka.actor.ActorSystem
+import akka.actor.{ ActorRefFactory, ActorSystem }
 import net.codingwell.scalaguice.ScalaModule
 
 import scala.concurrent.ExecutionContext
 
-/** Module that binds an ActorSystem and its associated ExecutionContext.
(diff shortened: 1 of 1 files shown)


codeviking:
🍌 🍎 🍓 🍒 

LGTM.  Merge on 🍏

codeviking:
Side note, does an `ActorRefFactor` implicitly satisfy the an `ActorSystem` but in a fashion that `compresssResponseIfRequested` doesn't work?

jkinkead:
I have no idea what the words you just typed mean.

codeviking:
Sorry, what does this fix?

jkinkead:
It means that you can have a dependency on the interface you need, instead of a superset of that interface. Just a smaller API surface; more precise dependencies. :)

codeviking:
Ok, so this just makes that class accessible for use elsewhere (I assume future PRs intend to use it)?

jkinkead:
Yeah, I was trying to inject [a class that was using an implicit `ActorSystem` as an `ActorRefFactory`](https://github.com/allenai/aristo/blob/bd6c11c987dc69237daf6923572aa215493195ac/evaluation/web/src/main/scala/org/allenai/ari/evaluationframework/routing/EvaluationResultRouting.scala#L22), and it failed because this type doesn't have a binding, even though the `ActorSystem` does.

PR #193, opened 2016-06-22
Make enum serialization more automatic.

jkinkead:
This uses the `toString` as the serialization value (by default). This reduces the boilerplate when specifying enum values, and prevents errors introduced by accidentally using the same `id` string twice, like in:
```scala
case object MyValue1 extends MyEnum("value1")
case object MyValue2 extends MyEnum("value1") // whoops!
```
The one thing that I considered, but didn't include, is adding a self-type of `Product` to the `Enum` mixin. It's not technically required; `toString` exists on `Any`. However, the self-type encourages you to use `case object`s for the enum values (you can technically use any `case class` or even `Tuple`, but that requires some serious effort). It also reduces the potential for bugs in migrating from the old version.

On the downside, it requires a self-type (or extending `Product`) on the abstract parent class:
```scala
// Using a self-type:
abstract class Enum[E <: Enum[E]] { self: Product =>
  // etc.
}
// Additional boilerplate:
sealed abstract class MyEnum extends Enum[MyEnum] { self: Product => }
```
Mostly I worry that the protection it gives is basically nil, but it adds some annoyance to usage. If there were a `Product0` or similar class that only `case object`s extended I would use that, but there isn't AFAICT.

### DIFF
--- a/core/src/main/scala/org/allenai/common/Enum.scala
+++ b/core/src/main/scala/org/allenai/common/Enum.scala
@@ -1,32 +1,33 @@
 package org.allenai.common
 
-import spray.json._
-import spray.json.DefaultJsonProtocol._
+import spray.json.{ deserializationError, JsString, JsValue, RootJsonFormat }
 
-/** Enumeration implementation that supports automatic Spray JSON serialization as JsString(id).
-  *
-  * @param id  value used for lookup and JSON serialization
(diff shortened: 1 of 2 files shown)


codeviking:
LGTM.  🚢  it!

jkinkead:
Great, merging!

I'll include migration instructions in the release notes.

PR #194, opened 2016-06-23
Update ElasticSearch indexing to 2.3.3.

ColinArenz:


indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
```diff
@@ -61,10 +61,7 @@ class BuildCorpusIndex(config: Config) extends Logging {
       val createIndexRequestBuilder: CreateIndexRequestBuilder =
         esClient.admin().indices().prepareCreate(indexName)
 
-      // Optimize settings for indexing
```
ColinArenz: We don't want to optimize for indexing in this way.  Shard count can't be changed after creating an index, and the new cluster is optimized for load balancing across a few nodes.  Current defaults are 5 shards, 1 replica, which provides pretty good load balancing and allows us to add/remove/replace nodes on a whim without snapshot/restore.

ColinArenz:
@OyvindTafjord FYI

ColinArenz:
Switching over to Oyvind for review due to vacation schedule.

OyvindTafjord:
LGTM! Merging.

PR #197, opened 2016-07-11
Add Bing client query API.

ColinArenz:
This is a straight copy and slight cleanup of the prototype from @joelgrus .  I left it with okHttp instead of sticking with spray-client, as I think we have a broad desire to move away from spray.  This does create a minor inconsistency, but hopefully a short-lived one.

### DIFF
--- a/webapp/src/main/scala/org/allenai/common/webapp/BingClient.scala
+++ b/webapp/src/main/scala/org/allenai/common/webapp/BingClient.scala
@@ -83,7 +83,7 @@ class BingClient(apiKey: String) {
     * @param top number of desired results for each query
     * @return a sequence of futures mapping each result
     */
-  def bulkQuery(queries: Seq[String], top: Int = 10) = queries.map {
+  def bulkQuery(queries: Seq[String], top: Int = 10): Seq[Future[Seq[BingResult]]] = queries.map {
     q => Future { query(q, top = top) }
   }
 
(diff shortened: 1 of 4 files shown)


webapp/src/main/scala/org/allenai/common/webapp/BingClient.scala
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+  }
+
+  /** Execute several queries in parallel, running the supplied callback on each individual result.
+    * @param queries sequence of queries to submit
+    * @param top number of desired results for each query
+    * @return a sequence of futures mapping each result
+    */
+  def bulkQuery(queries: Seq[String], top: Int = 10) = queries.map {
```
joelgrus: I would add return type annotations on these, just to make it clearer

webapp/src/main/scala/org/allenai/common/webapp/BingClient.scala
```diff
@@ -0,0 +1,99 @@
...
+    q => Future { query(q, top = top) }
+  }
+
+  /** A helper function to get a string out of a JsObject
+    * @param key name of the field
+    * @param jsMap the map of Json fields
+    * @return the string, or None if unavailable
+    */
+  private def getString(key: String, jsMap: Map[String, JsValue]) = jsMap.get(key).flatMap {
```
joelgrus: similarly, I wrote this function, and I had to stare at it for a minute to remind me what I did, it would help if the return type was specified :)

joelgrus:
:+1: looks good, just if you can add the two return type annotations

joelgrus:
:+1:

PR #199, opened 2016-07-28
Index Quizlet and Studystack

sbhaktha:
@ColinArenz : Each of these corpora was split into two -- one containing `question:|:answer` pairs and the other containing `term:|:definition` pairs, one per line. This code indexes the corpora with two fields instead of a single `text` field like we do for our other Aristo corpora.

### DIFF
--- a/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
+++ b/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
@@ -75,6 +75,67 @@ org.allenai.common.indexing.paragraph_base: ${org.allenai.common.indexing.base} 
   }
 }
 
+org.allenai.common.indexing.qa_base: ${org.allenai.common.indexing.base} {
+  elasticSearch: {
+    indexType: "sentence"
+    mapping: {
+      sentence: {
+        dynamic: false,
+        properties: {
+          question: {
+            type: "string",
+            analyzer: "snowball", 
+            fields: {
+              raw: {type: "string", index: "not_analyzed"}
+            }
+          }
+          answer: {
+            type: "string",
+            analyzer: "snowball",
+            fields: {
+              raw: {type: "string", index: "not_analyzed"}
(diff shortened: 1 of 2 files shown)


indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -75,6 +75,67 @@ org.allenai.common.indexing.paragraph_base: ${org.allenai.common.indexing.base}
   }
 }
 
+org.allenai.common.indexing.qa_base: ${org.allenai.common.indexing.base} {
+  elasticSearch: {
+    indexType: "sentence"
```
ColinArenz: I'd imagine you'd want to change the type to "question" or something similar?  That would allow this type of document to coexist with existing corpora in the same index, if that's something we ever want to do.

indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
```diff
@@ -238,21 +238,58 @@ class BuildCorpusIndex(config: Config) extends Logging {
...
     bulkProcessor: BulkProcessor
   ): Unit = {
-    val request = new IndexRequest(indexName, indexType).source(jsonBuilder().startObject()
-      .field("text", segment.trim)
-      .field("source", source + "_" + segmentIndex.toString)
-      .endObject())
-    bulkProcessor.add(request)
+    // Helper Function
+    def breakQAline(line: String): Option[(String, String)] = {
+      line.split(""":\|:""") match {
```
ColinArenz: You can skip the double-trim if you want to: `line.split(...).map(_.trim) match { ... }`

ColinArenz:
LGTM.  I left a couple of comments, but they're not blocking.  Merge whenever you're ready. :)

sbhaktha:
Addressed both comments. Merging!

PR #200, opened 2016-07-28
make caching constructor protected.

danyaljj:


jkinkead:
I'd just make this public - LGTM either way!

PR #202, opened 2016-07-29
Small method adding to the Jedis wrapper 

danyaljj:
Adding a new method in JedisWrapper for getting keys based on some some pattern 

@ColinArenz @jkinkead for either of one you have time/energy.


ColinArenz:
LGTM!

PR #205, opened 2016-08-08
adding a new constructor fro JedisClient 

danyaljj:
To be used in https://github.com/allenai/aristo/pull/2468

### DIFF
--- a/cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
+++ b/cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
@@ -25,8 +25,8 @@ object JsonQueryCache {
   }
 
   def apply[V](
-    hostname: String,
     clientPrefix: String,
+    hostname: String,
     port: Int = Protocol.DEFAULT_PORT,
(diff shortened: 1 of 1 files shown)


cache/src/main/scala/org/allenai/common/cache/JsonQueryCache.scala
```diff
@@ -21,11 +21,21 @@ object JsonQueryCache {
     val port: Int = config.get[Int]("port") getOrElse Protocol.DEFAULT_PORT
     val timeoutMillis: Int = config.get[Int]("timeoutMillis") getOrElse Protocol.DEFAULT_TIMEOUT
 
+    apply[V](hostname, clientPrefix, port, timeoutMillis)
+  }
+
+  def apply[V](
+    hostname: String,
+    clientPrefix: String,
```
ColinArenz: Bit of a nitpick, but this should probably be the first parameter to match with the other constructor.  The grouping is conceptually `(clientPrefix, (hostname, port, timeoutMillis))`.

ColinArenz:
One minor comment.  Otherwise LGTM.

danyaljj:
Addressed the comment. Will merge on green.

PR #209, opened 2016-09-14
Clean up ElasticSearch index mappings

ColinArenz:
This change reduces copy-paste index mappings in `indexing.conf` and adds a date tag to index names to allow for reindexing an existing corpus.

### DIFF
--- a/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
+++ b/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
@@ -1,24 +1,24 @@
 include "mappingHelpers.conf"
 
-org.allenai.common.indexing {
-  stoplist {
-    group: "org.allenai.nlp.resources"
-    name: "smart_stopwords_2.txt"
-    version: 1
-  }
+org.allenai.common.indexing.stoplist {
+  group: "org.allenai.nlp.resources"
+  name: "smart_stopwords_2.txt"
+  version: 1
 }
 
 org.allenai.common.indexing.base {
   elasticSearch {
-    // Specify the right clustername. It usually defaults to `elasticsearch` but
-    // you can verify by using the command:
-    // curl -XGET 'http://localhost:9200/_cluster/health?pretty=true'.
-    // For Aristo Solvers, the hostIp should be changed to `aristo-es.dev.ai2`
-    // with clusterName `aristo-es`.
-    clusterName: "elasticsearch"
-    hostIp: "127.0.0.1"
+    hostAddress: "127.0.0.1"
     hostPort: 9300
+
+    // Your cluster name can be obtained with the following command.
+    // curl -XGET 'http://<hostAddress>:9200/_cluster/health?pretty=true'.
+    clusterName: "elasticsearch"
     indexName: "lucene"
+
+    // Aristo should override the following values:
+    //clusterName: "aristo-es"
+    //hostAddress: "aristo-es.dev.ai2"
   }
 
   buildIndexOptions {
--- a/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
+++ b/indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
@@ -1,19 +1,19 @@
-org.allenai.common.indexing: {
+include "mappingHelpers.conf"
 
-  stoplist: {
+org.allenai.common.indexing {
+  stoplist {
     group: "org.allenai.nlp.resources"
     name: "smart_stopwords_2.txt"
     version: 1
   }
-
 }
 
-org.allenai.common.indexing.base: {
-  elasticSearch: {
+org.allenai.common.indexing.base {
+  elasticSearch {
     // Specify the right clustername. It usually defaults to `elasticsearch` but
     // you can verify by using the command:
     // curl -XGET 'http://localhost:9200/_cluster/health?pretty=true'.
-    // For Aristo Solvers, the hostIp should be changed to `aristo-es1.dev.ai2`
+    // For Aristo Solvers, the hostIp should be changed to `aristo-es.dev.ai2`
     // with clusterName `aristo-es`.
     clusterName: "elasticsearch"
     hostIp: "127.0.0.1"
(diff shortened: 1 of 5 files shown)


indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -1,136 +1,88 @@
...
+  buildIndexOptions {
     buildFromScratch: true
     dumpFolder: "common/src/main/resources/org/allenai/ari/indexing/dump"
   }
 }
 
 org.allenai.common.indexing.sentence_base: ${org.allenai.common.indexing.base} {
-  elasticSearch: {
+  elasticSearch {
     indexType: "sentence"
```
ColinArenz: These `indexType` declarations are redundant, but I can't quite remove them yet.

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -1,136 +1,88 @@
-org.allenai.common.indexing: {
+include "mappingHelpers.conf"
 
-  stoplist: {
+org.allenai.common.indexing {
+  stoplist {
```
aimichal: Since there's only one entry here, you could name the top-level with suffix `.stoplist` and remove one level of indentation.

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -1,136 +1,88 @@
...
+  elasticSearch {
     // Specify the right clustername. It usually defaults to `elasticsearch` but
     // you can verify by using the command:
     // curl -XGET 'http://localhost:9200/_cluster/health?pretty=true'.
-    // For Aristo Solvers, the hostIp should be changed to `aristo-es1.dev.ai2`
+    // For Aristo Solvers, the hostIp should be changed to `aristo-es.dev.ai2`
     // with clusterName `aristo-es`.
     clusterName: "elasticsearch"
     hostIp: "127.0.0.1"
     hostPort: 9300
```
aimichal: The curl command says to use port 9200, but hostPort is 9300. Intentional?

Also, of less concern to me:

1. `hostIp` is a hostname or an IP, so `host` is probably a better name for this field.
2. The comment `Specify the right clustername` is quite removed from the `clusterName`  field. I suggest ordering the comments/fields like this:

```
// For Aristo Solvers, ...
hostIp: ...
hostPort: ...

// Specify the right clustername. ...
clusterName: "..."
```

indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
```diff
@@ -36,7 +38,13 @@ class BuildCorpusIndex(config: Config) extends Logging {
 
   /** Get Index Name and Index Type. */
   val esConfig: Config = config[Config]("elasticSearch")
-  val indexName: String = esConfig[String]("indexName")
+  val indexName: String = {
+    // The index name must be appended with the current date.
+    val name = esConfig[String]("indexName")
+    val dateFormat = new SimpleDateFormat("yyyy-MM-dd")
+    name + "-" + dateFormat.format(Calendar.getInstance().getTime())
```
aimichal: Not sure how important this is, but does this choose a time in UTC or local system time?

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
ColinArenz: 1. 9300 is the binary API, while 9200 uses JSON.
2. I don't fully understand the second point.  This entire commend applies to `hostAddress` (renamed) and `clusterName`.  I'll reorganize this a little to clarify.

indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
ColinArenz: This is just a light-weight disambiguation and we don't expect to reindex twice in the same day.  The date itself is arbitrary, so plus or minus 8 hours shouldn't matter.

aimichal: Is it possible that I'd start this in the morning, cancel it, and start over all in the same day? Would that erase the partially populated index first or supplement it?

I think that adding hour:minute:second to the name would remove this ambiguity.

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
```diff
@@ -1,24 +1,24 @@
...
-    hostIp: "127.0.0.1"
+    hostAddress: "127.0.0.1"
     hostPort: 9300
+
+    // Your cluster name can be obtained with the following command.
+    // curl -XGET 'http://<hostAddress>:9200/_cluster/health?pretty=true'.
+    clusterName: "elasticsearch"
     indexName: "lucene"
+
+    // Aristo should override the following values:
```
aimichal: "override with the following values"?

indexing/src/main/resources/org/allenai/common/indexing/indexing.conf
aimichal: It's weird to me that we've got an explicit config setting for `hostPort: 9300` and reference an implicit port number `9200` a few lines later. Would it make sense to rename `hostPort` to something like `transportPortNumber`?

Reading through https://www.elastic.co/guide/en/elasticsearch/guide/current/_talking_to_elasticsearch.html it sounds like `9200` and `9300` are default values -- is `hostPort: 9300` needed?

ColinArenz: I'd personally prefer to remove it until there's a strong need to configure the value.  All of our ES nodes have always used the default ports.

indexing/src/main/scala/org/allenai/common/indexing/BuildCorpusIndex.scala
ColinArenz: At current writing, that workflow would fail on the second attempt due to the index already existing.  It needs to be deleted manually.  I have a long-term task to rework this entire tool (which would delete a failed index), but I want to keep this change minimally invasive.

Hour/minute/second is a bit noisy, and I'd actually like to discourage extremely frequent (e.g. same day) reindexing, since we need to keep copies of any index that's been used for repeatability.

PR #211, opened 2016-10-20
Indexing CK-12 flexbooks

sbhaktha:
Also small change to not segment lines for q-a and term-def corpora and corpora specified as "sentence per line".


ColinArenz:
LGTM.

PR #212, opened 2016-11-14
update BingClient to bing v5 api

joelgrus:
main changes:

* update url to api.cognitive.microsoft.com
* source type becomes response filter (I don't think anyone was explicitly setting this)
* various other query changes
* results now at `json["webPages"]["value"]`
* ID -> id (I don't think it's exactly the same field either, I think the old one is deprecated)
* Url -> url, and is a bing redirect we need to get the actual url out of
* Title -> name
* Description -> snippet

see https://msdn.microsoft.com/en-US/library/mt707570.aspx#web for details

you also need a new v5 subscription key to use this, you can get it from techops, it's called aristo-bing-client


joelgrus:
FYI @rodneykinney

PR #215, opened 2017-01-17
Fix the handling of sequences of non-Config values.

jkinkead:
This works around both a weird API issue in Typesafe Config (the inability to test the type of a list) and some Scala-Java compatibility weirdness (the lack of a proper type parameter for boxed primitives).

Note that the current code throws an exception if provided with a config like:
```scala
value = [ "foo", "bar" ]
```


jkinkead:
The APIs are ugly, and having to do what is effectively chained try-catch blocks is going to be ugly either way, given the implicits required.

This is some spare-time hacking to fix bugs in `ConfigModule` picked up when playing around with it for the online project. Nothing related to key tasks, just potential code-debt cleanup in the future.

codeviking:
Great, thanks for the explanation!

PR #216, opened 2017-01-18
Release v1.4.7

jkinkead:
I'm letting you review this, since you just ported OpenIE to this release process.

This release contains bugfixes for the `ConfigModule`, some of which have hit researchers on Aristo.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "1.4.7-SNAPSHOT"
+version in ThisBuild := "1.4.7"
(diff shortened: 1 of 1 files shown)


aimichal:
This change by itself looks harmless, but the published changes are many going back to September:

https://github.com/allenai/common/compare/3a4a1bd3b4b8654032d2bc4ae1cb37cb7089a724...810e606db688bc2d9d90f78a3bd4ec2018be8794

I'm sure that's all good, right..? :-)

(As an aside, I'm shocked at how hard it was to generate the above URL; is there an easy way to do this?)

jkinkead:
Yep, I previewed the changes already. They should be harmless.

I don't know how you generated the URL. The most direct way I know is to append `/compare` to the base URL, then search up the two items you want to compare (the tags `v1.4.6` and `v1.4.7` in this case).

PR #218, opened 2017-01-19
Restore serialization when a constructor argument exists.

jkinkead:
Add a unit test to cover this case.

This fixes a backwards-compatibility issue - I'm restoring a feature I broke.

### DIFF
--- a/core/src/test/scala/org/allenai/common/EnumSpec.scala
+++ b/core/src/test/scala/org/allenai/common/EnumSpec.scala
@@ -40,6 +40,22 @@ class EnumSpec extends UnitSpec {
     }
   }
 
+  "Java serialization" should "work with no constructor argument" in {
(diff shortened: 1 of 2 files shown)


core/src/test/scala/org/allenai/common/EnumSpec.scala
```diff
@@ -40,8 +40,8 @@ class EnumSpec extends UnitSpec {
     }
   }
 
-  "Java serialization" should "work" in {
-    FakeEnum.all foreach { enum =>
+  it should "work with a constructor argument" in {
```
jkinkead: I just realized I accidentally deleted a test here - there's supposed to be two `Java serialization` tests, but I accidentally deleted one.

I'm adding it back in & merging. :)

PR #220, opened 2017-02-02
Fix bug injecting nested lists.

jkinkead:
Here's the root cause of the S2 bug.

I hate unit-tested, vetted code that still has bugs.


jkinkead:
Since this should mostly fail to bind anything, it should be pretty dang obvious. I also don't think anyone who's using this (common-guice) has updated since I introduced this bug.

vimsick:
:+1:

PR #221, opened 2017-02-02
Release version 1.4.9

jkinkead:
Don't squash-merge.


jkinkead:
The tagged release won't show up in the git history if you squash (the tag is attached to that first commit). It also breaks CI's ability to publish this.

vimsick:
:+1:

PR #226, opened 2020-06-05
cross publish to 2.13

bbstilson:
This PR makes some tweaks to the common code so that we can publish the common core for 2.13. Some things I decided to delete, specifically: `Interval` and `ScalaUtils`. They are no longer used.

There are also several changes that are non-functional changes but fix a lot of warnings (mostly unused imports and adding types and parens where appropriate).

Finally, my IDE took over and formatted a lot of stuff, so please ignore the whitespace changes 😓 

I've run `+test` locally and everything is 🍏 

I also ran `publishLocal` after forcing the scala version 2.13.2, and set up a dummy project locally, and confirmed that it works.


build.sbt
```diff
@@ -50,32 +51,43 @@ lazy val buildSettings = Seq(
...
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, x)) if x == 11 || x == 12 => {
+        Seq(file(sourceDirectory.value.getPath + "/main/scala-2.11-2.12"))
+      }
+      case Some((2, x)) if x == 13 => {
+        Seq(file(sourceDirectory.value.getPath + "/main/scala-2.13"))
+      }
+      case _ => Seq.empty // dotty support would go here
+    }
+  }
```
bbstilson, line 64: This enables sbt to read different source directories.

core/src/main/scala-2.13/org/allenai/common/ScalaUtils.scala
```diff
@@ -0,0 +1,3 @@
+package org.allenai.common
+
+object ScalaUtils {}
```
bbstilson, line 3: This is specifically for 2.13. I'm not sure the method in here was even used, but I don't think we'll need it for 2.13. Specifically because `mapValues` is deprecated.

bbstilson:
Do you think it's safe to delete Interval? I did a grep of several main s2 repos and couldn't find anything.

schmmd:
I love `Interval`.  I think I wrote it... 😢  @bbstilson

bbstilson:
It will live forever in our (and git's) memory. ❤️

PR #232, opened 2021-04-12
drop spray patch

bbstilson:


regan-huff (approved):
certified and approved by the Spray Stays The Way It Is Committee

